PR #24, opened 2016-09-30
Scan Routines

felgonzalezh:
Scan routines were implemented in GEMSupervisor Applications.
It will send the scan parameter through GEMApplication to Manager Applications.
Manager Applications were modified in order to set the scanparameters into OH, GLIB and AMC13
AMC13Manager was modified in order to count the triggers and to manage the different triggers sources of the scans.


jsturdy:
Superseded by #25

PR #25, opened 2016-10-04
Merged in scan updates and cleaned up some code

jsturdy:
supersedes #24


jsturdy:
* Scans are working from the supervisor and manager applications
* Some bug still exists somewhere in the FSM that causes a seg fault sometimes
 * Haven't been able to track down precisely what's happening, but it doesn't impact normal data taking
 * Seems to have to do with frequent changes to the FSM state

PR #37, opened 2016-11-07
Tried to clean up python query

jsturdy:
The logic seemed to get broken so I tried to add pass statements.
I think I have managed to get it to be identical to what was done before.
Can you check this?

### DIFF
--- a/setup/scripts/python/query.py
+++ b/setup/scripts/python/query.py
@@ -13,78 +13,89 @@ def configure_db(station="TIF",setuptype="teststand",runperiod="2016T"):
     zlist = zip(amc_list, geb_list)
     a_list = []
     for amcN, gtx_list in zlist:
-      print "Trying to connect to AMC # %s\n" %(amcN)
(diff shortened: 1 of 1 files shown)


mexanick:
I don't understand why you have changed again the whole indentation scheme. I thought we agree to use **2 spaces as a tab and convert all tabs to spaces**
Let me (and other developers) know how it should be.

jsturdy:
I did the change because that's how my emacs was set up for python (it doesn't automatically indent python files when opening them)
However, some change was introduced (almost definitely my fault) and making sure everything was indented nicely (for my editor) seemed to be the quickest way to ensure this on my end.
As long as the pass statements are correctly placed, this doesn't matter.
Updated the commit with 2 spaces, but my only concern is that the file didn't get a change introduced during these steps.  Can you confirm this?

mexanick:
Looks good to me. Is there anything you want to check or I can merge the pull request?

jsturdy:
Nope!  Please go ahead and merge.

Thanks!

PR #99, opened 2017-03-26
Remove setting of ContReg3

jsturdy:
Removed setting of ```ContReg3``` from ```biasAllVFATs``` as this control register contains the setting for the trimDAC range, and will vary from chip to chip


jsturdy:
Added a ```wrappers``` utility module containing for now, just a wrapper around ```subprocess.call``` that is used regularly

PR #101, opened 2017-04-05
Add envCheck to gempython.utils.wrappers

cbravo135:


gempython/utils/wrappers.py
```diff
@@ -30,3 +30,11 @@ def runCommand(cmd,log=None):
         sys.stdout.flush()
         pass
     return returncode
+
+def envCheck(envVar):
+    import os
+    if os.getenv(envVar) == None or os.getenv(envVar) == '':
+        print 'You must source the environment properly, %s is not set'%envVar
+        exit(0)
```
jsturdy: I would take this and handle it in the calling code (```return False``` here)

jsturdy: could also raise an exception that is caught/handled in the calling code

PR #112, opened 2017-06-14
Version for p5

jsturdy:


jsturdy:
* ```useLocalDBInstance``` is a parameter of the ```GEMSupervisor```, the same as the other parameters that are passed in via the xml config file
* Regarding the other points, the file ```gemsupervisor.xml``` is only an example, updated to include the necessary new parameters, there will be a set of configuration files released for each setup separately in due time

mexanick:
I didn't find this `useLocalDBInstance` in the xml example. Probably better to have it there.

jsturdy:
* [```UseLocalDB``` in the xml file](https://github.com/jsturdy/cmsgemos/blob/generic-amc-release-v3/gemsupervisor/xml/gemsupervisor.xml#L138)
* [mapping here](https://github.com/jsturdy/cmsgemos/blob/generic-amc-release-v3/gemsupervisor/src/common/GEMSupervisor.cc#L77)

PR #120, opened 2017-06-20
added CONTRIBUTING.md

jsturdy:


jsturdy:
Also add templates for issues and pull requests, hopefully so that they are picked up by github, but have to understand how to implement that

Issue #117, opened 2017-06-17
Force error on failed SOAP transitions


jsturdy:
Currently SOAP commands sent to TCDS applications which can't reach their endpoint result in an application crash.
This needs to be changed to an error transition

jsturdy:
More info:
```
Unexpected exception in toolbox::task thread
terminate called after throwing an instance of 'gem::utils::exception::SOAPException'
  what():  Command Halt failed [Failed to post SOAP message]
Process [30789] on node gem904daq01.cern.ch terminates abnormally at 2017-06-21T14:44:16.083770+02:00. Caught signal 6 at address [0x43b100007845]
Stacktrace follows:
/opt/xdaq/lib/libtoolbox.so(_ZN7toolbox10stacktraceEiRSo+0x44) [0x7f17ec914464]
/opt/xdaq/lib/libtoolbox.so(_ZN7toolbox18signalSEGVCallbackEiP7siginfoPv+0x262) [0x7f17ec9150f2]
/lib64/libpthread.so.0() [0x3f0f20f7e0]
```
For some reason (yet to be fully understood) the throw is not being caught and is instead triggering a ```toolbox::Task::Task::unexpected_taskhandler``` which subsequently calls ```abort```

PR #121, opened 2017-06-21
hotfix for failed SOAP calls crashing application

jsturdy:
* Changed throw specifier so that ```std::unexpected``` is not called,
* fixes cms-gem-daq-project/cmsgemos#117


jsturdy:
I had thought that I was throwing an exception ```gem::utils::exception::SOAPException``` that was a subset of the ```throw (gem::utils::exception::Exception)``` specifier, but it wasn't...
If a function ```throw```/```raise```/```rethrow```/```reraise```s an exception that is not within the ```throw``` specifier, ```std::unexpected``` gets called, and in the ```xdaq``` code, the handler basically calls an ```abort```
I think this issue exists in other code as well, so it will need to be reviewed.

Issue #127, opened 2017-07-10
Latency scan monitoring


mexanick:
## Brief summary of issue
Latency scan doesn't provide any monitoring capabilities to see how the scan is progressing

### Types of issue

- [ ] Feature request: add certain monitoring printouts to the log file

## Context (for feature requests)

I'm going to add following information:
- OH CRC counters for each VFAT at the beginning of the scan and after it finishes- this might affect the number of triggers we take since the data with failed CRC may be discarded. _QUESTION_: do we want to add this printout (24 values per chamber) for all the scans or only for Latency?
- Number of L1As reported by AMC13, CTP7 and OH (each 1000 L1As send from AMC13) -- _only for Latency scan_

Do I miss anything? Any other suggestions?

cbravo135:
Does a CRC error counter for each VFAT already exist or is it just one which is a sum over all VFATs?

cbravo135:
I also think we should add a printout of the OH CRC counters to optohybrid_info_uhal.py in cmsgemos.

mexanick:
These are per-VFAT counters. Also they could be reset individually at the beginning of the run

cbravo135:
Yes, lets reset them at the beginning of the scan and read them into a printout at the end.

jsturdy:
Regarding the question 
> Do we want to add this printout (24 values per chamber) for all the scans or only for Latency:

it would only make sense for a scan that is using the tracking data, as that's the only case where it would directly tell us anything
* Now that I understand your request better, the next question is how often do you really want this progress information to be logged?
  * Do you really want it to be fixed to be per/1000? I'd say that as a start it could be once per N triggers requested per point
* Still better would be adding a progress counter to the firmware module itself, that can be read from the status register, reporting the current scan value.
  * This could be added to the additional register I mention below as there would be 8 bits free anyway
* What is the purpose for adding CRC printing?
  * If it's what I think (debugging issues we've been seeing), then I would add it in the firmware and put it in the status register (24 bits in a new status register) with a 1 signifying that a given VFAT had a CRC error during the scan
  * Starting the scan automatically resets this to 0
  * The status register currently has 24 bits which are used to report which VFATs have data that are available for readout.  This is not (at the moment) directly used and could be repurposed for this new status bit, but better would be to add another register for this additional status check.
  * It would not preclude from printing out the CRC error counts, but would be a nice succinct way to summarize the information that I think @mexanick is hoping to add.

[...]

PR #129, opened 2017-07-11
Adding modifications according to issue #127

mexanick:
Adding modifications according to issue #127 (enhancing debug printouts and monitoring of ultraLatency calibration routine
## Description

* Adding L1A counters getters
* Adding incorrect CRC counter reset at the scan module configuration stage
* Adding printouts to getUltraScanResults in case of Latency scan
### Types of changes

## Motivation and Context

https://github.com/cms-gem-daq-project/cmsgemos/issues/127

## How Has This Been Tested?

Requires testing 
## Checklist:

### DIFF
--- a/gempython/tools/amc_user_functions_uhal.py
+++ b/gempython/tools/amc_user_functions_uhal.py
@@ -175,7 +175,3 @@ def enableL1A(device):
 
 def getAMCL1Acount(device):
     return readRegister(device, "GEM_AMC.TTC.COUNTERS.L1A")
-                                                                                          
-def getLinkL1Acount(device,link):
-    return readRegister(device, "GEM_AMC.OH.OH%s.COUNTERS.T1.SENT.L1A"%(link))
-
--- a/gempython/tools/amc_user_functions_uhal.py
+++ b/gempython/tools/amc_user_functions_uhal.py
@@ -173,5 +173,5 @@ def blockL1A(device):
 def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
 
-def getAMCL1Acount(device):
+def getL1ACount(device):
     return readRegister(device, "GEM_AMC.TTC.CMD_COUNTERS.L1A")
--- a/gempython/tools/amc_user_functions_uhal.py
+++ b/gempython/tools/amc_user_functions_uhal.py
@@ -172,3 +172,10 @@ def blockL1A(device):
 
 def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
+
+def getAMCL1Acount(device):
+    return readRegister(device, "GEM_AMC.TTC.L1A_ID")
+                                                                                          
+def getLinkL1Acount(device,link):
+    return readRegister(device, "GEM_AMC.DAQ.OH%s.COUNTERS.EVN"%(link))
+
--- a/gempython/tools/amc_user_functions_uhal.py
+++ b/gempython/tools/amc_user_functions_uhal.py
@@ -177,5 +177,5 @@ def getAMCL1Acount(device):
     return readRegister(device, "GEM_AMC.TTC.L1A_ID")
                                                                                           
 def getLinkL1Acount(device,link):
-    return readRegister(device, "GEM_AMC.DAQ.OH%s.COUNTERS.EVN"%(link))
+    return readRegister(device, "GEM_AMC.OH.OH0.COUNTERS.T1.SENT.L1A"%(link))
 
--- a/gempython/tools/amc_user_functions_uhal.py
+++ b/gempython/tools/amc_user_functions_uhal.py
@@ -174,4 +174,4 @@ def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
 
 def getAMCL1Acount(device):
-    return readRegister(device, "GEM_AMC.TTC.COUNTERS.L1A")
+    return readRegister(device, "GEM_AMC.TTC.CMD_COUNTERS.L1A")
--- a/gempython/tools/amc_user_functions_uhal.py
+++ b/gempython/tools/amc_user_functions_uhal.py
@@ -174,8 +174,8 @@ def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
 
 def getAMCL1Acount(device):
-    return readRegister(device, "GEM_AMC.TTC.L1A_ID")
+    return readRegister(device, "GEM_AMC.TTC.COUNTERS.L1A")
                                                                                           
 def getLinkL1Acount(device,link):
-    return readRegister(device, "GEM_AMC.OH.OH0.COUNTERS.T1.SENT.L1A"%(link))
+    return readRegister(device, "GEM_AMC.OH.OH%s.COUNTERS.T1.SENT.L1A"%(link))
 
--- a/gempython/tools/optohybrid_user_functions_uhal.py
+++ b/gempython/tools/optohybrid_user_functions_uhal.py
@@ -703,10 +703,16 @@ def getScanResults(device, gtx, numpoints, debug=False):
 
     return results
 
-def getUltraScanResults(device, gtx, numpoints, debug=False, numtrigs=1000, isLatency=False):
+def getUltraScanResults(device, gtx, numpoints, debug=False):
     scanBase = "GEM_AMC.OH.OH%d.ScanController.ULTRA"%(gtx)
     ohnL1A_0 = getLinkL1Acount(device,gtx)
     ohnL1A = getLinkL1Acount(device,gtx)
+    numtrigs = readRegister(device,"%s.NTRIGS"%(scanBase))
+    if (readRegister(device,"%s.MODE"%(scanBase))==2):
+        isLatency = True
+        print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
+    else:
+        isLatency = False    
     while (readRegister(device,"%s.MONITOR.STATUS"%(scanBase)) > 0):
         msg = "%s: Ultra scan still running (0x%x), not returning results"%(device,
                                                                             readRegister(device,"%s.MONITOR.STATUS"%(scanBase)))
--- a/gempython/tools/optohybrid_user_functions_uhal.py
+++ b/gempython/tools/optohybrid_user_functions_uhal.py
@@ -595,8 +595,11 @@ def configureScanModule(device, gtx, mode, vfat, channel=0,
 
     writeRegisterList(device,regList)
     
-    regList = dict.fromkeys(["GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d.Reset"%(gtx,i) for i in range(24)],1)
-    writeRegisterList(device,regList)
+    #regList = dict.fromkeys(["GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d.Reset"%(gtx,i) for i in range(24)],1)
+    #writeRegisterList(device,regList)
+    # for some reason the code above doesn't work and triggers ipbus transaction errors... The code below works!
+    for i in range(24):
+      writeRegister(device,"GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d.Reset"%(gtx,i), 1)
 
     return
 
--- a/gempython/tools/optohybrid_user_functions_uhal.py
+++ b/gempython/tools/optohybrid_user_functions_uhal.py
@@ -807,5 +807,5 @@ def calculateLinkErrors(device,gtx,sampleTime):
         errorCounts[link] = [first,second]
     return errorCounts
                                                                                       
-def getLinkL1Acount(device,link):
+def getL1ACount(device,link):
     return readRegister(device, "GEM_AMC.OH.OH%s.COUNTERS.T1.SENT.L1A"%(link))
--- a/gempython/tools/optohybrid_user_functions_uhal.py
+++ b/gempython/tools/optohybrid_user_functions_uhal.py
@@ -594,6 +594,10 @@ def configureScanModule(device, gtx, mode, vfat, channel=0,
         pass
 
     writeRegisterList(device,regList)
+    
+    regList = dict.fromkeys(["GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d.Reset"%(gtx,i) for i in range(24)],1)
+    writeRegisterList(device,regList)
+
     return
 
 def printScanConfiguration(device,gtx,useUltra=False,debug=False):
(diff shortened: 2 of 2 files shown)


gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -696,12 +700,20 @@ def getScanResults(device, gtx, numpoints, debug=False):
...
     scanBase = "GEM_AMC.OH.OH%d.ScanController.ULTRA"%(gtx)
+    ohnL1A_0 = getLinkL1Acount(device,gtx)
+    ohnL1A = getLinkL1Acount(device,gtx)
     while (readRegister(device,"%s.MONITOR.STATUS"%(scanBase)) > 0):
         msg = "%s: Ultra scan still running (0x%x), not returning results"%(device,
                                                                             readRegister(device,"%s.MONITOR.STATUS"%(scanBase)))
         ohlogger.debug(colormsg(msg,logging.DEBUG))
+        if (isLatency):
+            if ((getLinkL1Acount(device,gtx)-ohnL1A) > numtrigs):
+                print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
```
cbravo135: We might want to discuss what exactly we want to have here, if anything. @jsturdy you have any comments on this part?

gempython/tools/amc_user_functions_uhal.py
```diff
@@ -172,3 +172,10 @@ def blockL1A(device):
 
 def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
+
+def getAMCL1Acount(device):
+    return readRegister(device, "GEM_AMC.TTC.L1A_ID")
+                                                                                          
+def getLinkL1Acount(device,link):
+    return readRegister(device, "GEM_AMC.DAQ.OH%s.COUNTERS.EVN"%(link))
+
```
jsturdy: These are not probably what you want.
* Do you want a count of the number of triggers sent to the VFATs?
  * `GEM_AMC.OH.OH0.COUNTERS.T1.SENT.L1A`
* Do you want the number of events built from VFAT data sent to the AMC?
  * Then what you have will do this, but not all running modes will enable the DAQ link and event building, so this is probably the wrong way to count.

gempython/tools/amc_user_functions_uhal.py
```diff
@@ -172,3 +172,10 @@ def blockL1A(device):
 
 def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
+
+def getAMCL1Acount(device):
+    return readRegister(device, "GEM_AMC.TTC.L1A_ID")
```
jsturdy: These are not probably what you want.
* Do you want a count of the number of triggers seen at the AMC?
  * `GEM_AMC.TTC.COUNTERS.L1A`
* Do you want the number of triggers seen since the DAQ module was reset?
  * Then what you have will do this, but not all running modes will enable the DAQ link and event building, so this is probably the wrong way to count.

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -594,6 +594,10 @@ def configureScanModule(device, gtx, mode, vfat, channel=0,
         pass
 
     writeRegisterList(device,regList)
+    
+    regList = dict.fromkeys(["GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d.Reset"%(gtx,i) for i in range(24)],1)
```
jsturdy: Why not move line 597 to 595 and append?

gempython/tools/optohybrid_user_functions_uhal.py
jsturdy: * You can get `numtrigs` from the scan config
* I'm be way more in favour of simply printing this out if `debug` is `True` rather than adding more arguments to the function
* Did you test that this behaves in a reasonable way?  The high rate behaviour is what I'm dubious of.

mexanick: > You can get `numtrigs` from the scan config
But this is not a class method - could you please elaborate how this can be implemented?
> I'm be way more in favour of simply printing this out if debug is True rather than adding more arguments to the function
We only want to print it for Latency scan, not for other scans. We can wrap it in debug==True, if you want, but to me that's only a progress-bar like indicator and doesn't require that.
> Did you test that this behaves in a reasonable way? The high rate behaviour is what I'm dubious of.
Not yet - I wanted to discuss first with you guys and then request a timeslot for testing at QC8

gempython/tools/amc_user_functions_uhal.py
mexanick: The counter I use shows the number of L1As since last EC0. I believe it doesn't interfere with DAQ

gempython/tools/amc_user_functions_uhal.py
mexanick: Agree. Change implemented

gempython/tools/optohybrid_user_functions_uhal.py
mexanick: for the moment for readability. But I can do it after the mods are properly tested

gempython/tools/amc_user_functions_uhal.py
jsturdy: It's not that it interferes with the DAQ, it's that it's completely coupled to using the DAQ module.
Nothing in the way the firmware scans are run does anything to this module.
* If it is reset when the AMC receives an EC0, this may not be what we want either, as the EC0 could come from upstream when we're not "running"
* As I understand it, this `L1A_ID` register is designed to be used in conjunction with other DAQ module registers, as setting up the AMC to take a "run" will initialize this module and reset everything as needed (something the firmware based scans won't do)
So the question that needs to be answered is: exactly what information is this function supposed to give?

gempython/tools/amc_user_functions_uhal.py
jsturdy: Change as implemented has hard coded `OH0`

gempython/tools/amc_user_functions_uhal.py
mexanick: ok, I'm going to change it to your suggestion. The goal is just to know how many L1As have been seen by CTP7

gempython/tools/amc_user_functions_uhal.py
cbravo135: link is an int but is being passed into the string with a %s and not a %i, is this ok to do?

mexanick: As far as I know it works (used a lot of times)

mexanick: This has been fixed already and this function migrated to optohybrid_user_functions

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -696,12 +703,20 @@ def getScanResults(device, gtx, numpoints, debug=False):
...
     scanBase = "GEM_AMC.OH.OH%d.ScanController.ULTRA"%(gtx)
+    ohnL1A_0 = getLinkL1Acount(device,gtx)
+    ohnL1A = getLinkL1Acount(device,gtx)
     while (readRegister(device,"%s.MONITOR.STATUS"%(scanBase)) > 0):
         msg = "%s: Ultra scan still running (0x%x), not returning results"%(device,
                                                                             readRegister(device,"%s.MONITOR.STATUS"%(scanBase)))
         ohlogger.debug(colormsg(msg,logging.DEBUG))
+        if (isLatency):
+            if ((getLinkL1Acount(device,gtx)-ohnL1A) > numtrigs):
+                print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
```
jsturdy: I agree that the printing should be somewhat dynamic.
I don't think it's very important to see an update at a fixed (in time) interval, the important thing is to simply see that the scan is progressing.  Some combination of the number of triggers per point and some number of updates during each point works, as does printing out at regular completion percentages.
You could consider something like [this](https://github.com/jsturdy/vfatqc-python-scripts/blob/dacscans/dacScan1.py#L7)
* You could print out something simple like:
```
<DAC name> <value>: <progress bar> | <total/partial completion percentage>
```
with a new line for each DAC value e.g.,
```
Latency 34: ====================> | 100%
Latency 35: ====================> | 100%
Latency 36: ============>         |  60%
```
or an overwritten line (where the progress bar is always moving up, but the latency value can change, as well as the percentage number, such that percentage is total, rather than per point):
```
Latency 36: ======>               | 37%
```
* Or something a little more fancy like (but requiring a bit more trickery):
```
<total progress bar> | <total completion percentage>
<DAC name> <value>: <DAC progress bar> | <DAC completion percentage>
```
e.g.,
```
Scan      : ===>                  |  23%
Latency 34: ====================> | 100%
Latency 35: ====================> | 100%
Latency 36: ============>         |  60%
```

gempython/tools/amc_user_functions_uhal.py
```diff
@@ -172,3 +172,6 @@ def blockL1A(device):
 
 def enableL1A(device):
     writeRegister(device, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x1)
+
+def getAMCL1Acount(device):
+    return readRegister(device, "GEM_AMC.TTC.CMD_COUNTERS.L1A")
```
mexanick: the problem here is that we don't have classes in python while we do in xDAQ. Also python doesn't support methods overloading. I can see two solutions, both of which I don't really like...
* I can make the functions for AMC and for OH with the same name and call them as module_name.function_name, but this might be ambigous
* I can set default link value to `-1` and use it for AMC, and if the link value provided, use it for OH - but this is also very artificial and won't make code very clear...

jsturdy: > I can make the functions for AMC and for OH with the same name and call them as module_name.function_name, but this might be ambigous

This is the "correct" way to do things, (the same as in `c` where you have to specify the namespace for conflicting function names.
You'd do something akin to:
```python
import <path to optohybrid_user_functions> as oh
import <path to amc_user_functions> as amc
...
amcTrigs = amc.getL1ACount(args)
ohTrigs = oh.getL1ACount(args)
```

The second option I reject outright

mexanick:
@jsturdy I implemented your requests, however this require additional testing (quite a lot of lines changed... could be typos). Let me know if you have any further comments before I can test it.

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -698,10 +705,24 @@ def getScanResults(device, gtx, numpoints, debug=False):
 
 def getUltraScanResults(device, gtx, numpoints, debug=False):
     scanBase = "GEM_AMC.OH.OH%d.ScanController.ULTRA"%(gtx)
+    ohnL1A_0 = getLinkL1Acount(device,gtx)
+    ohnL1A = getLinkL1Acount(device,gtx)
+    numtrigs = readRegister(device,"%s.NTRIGS"%(scanBase))
+    if (readRegister(device,"%s.MODE"%(scanBase))==2):
+        isLatency = True
+        print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
```
jsturdy: This still shows as `getLinkL1Acount`

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -698,10 +705,24 @@ def getScanResults(device, gtx, numpoints, debug=False):
...
+    else:
+        isLatency = False    
     while (readRegister(device,"%s.MONITOR.STATUS"%(scanBase)) > 0):
         msg = "%s: Ultra scan still running (0x%x), not returning results"%(device,
                                                                             readRegister(device,"%s.MONITOR.STATUS"%(scanBase)))
         ohlogger.debug(colormsg(msg,logging.DEBUG))
+        if (isLatency):
+            if ((getLinkL1Acount(device,gtx)-ohnL1A) > numtrigs):
+                print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
+                ohnL1A = getLinkL1Acount(device,gtx)
```
jsturdy: This still shows as `getLinkL1Acount`

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -698,10 +705,24 @@ def getScanResults(device, gtx, numpoints, debug=False):
...
+        isLatency = True
+        print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
+    else:
+        isLatency = False    
     while (readRegister(device,"%s.MONITOR.STATUS"%(scanBase)) > 0):
         msg = "%s: Ultra scan still running (0x%x), not returning results"%(device,
                                                                             readRegister(device,"%s.MONITOR.STATUS"%(scanBase)))
         ohlogger.debug(colormsg(msg,logging.DEBUG))
+        if (isLatency):
+            if ((getLinkL1Acount(device,gtx)-ohnL1A) > numtrigs):
```
jsturdy: This still shows as `getLinkL1Acount`

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -698,10 +705,24 @@ def getScanResults(device, gtx, numpoints, debug=False):
...
+                print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
```
jsturdy: This still shows as `getLinkL1Acount`

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -703,10 +703,16 @@ def getScanResults(device, gtx, numpoints, debug=False):
...
     ohnL1A = getLinkL1Acount(device,gtx)
+    numtrigs = readRegister(device,"%s.NTRIGS"%(scanBase))
+    if (readRegister(device,"%s.MODE"%(scanBase))==2):
+        isLatency = True
+        print "At link %s: %d/%d L1As processed, %d%% done" %(gtx, getLinkL1Acount(device,gtx)-ohnL1A_0, numpoints*numtrigs, (getLinkL1Acount(device,gtx)-ohnL1A_0)*100./(numpoints*numtrigs))
+    else:
+        isLatency = False    
     while (readRegister(device,"%s.MONITOR.STATUS"%(scanBase)) > 0):
         msg = "%s: Ultra scan still running (0x%x), not returning results"%(device,
                                                                             readRegister(device,"%s.MONITOR.STATUS"%(scanBase)))
```
cbravo135: This is only printing the extra information for latency scans, which is what we need now. We should discuss later how we would like to report this information in the case of other scans.

PR #141, opened 2017-09-12
Update vfat_info_uhal.py

jsturdy:
Remove hard-coded setting of `VThreshold1` from historical reasons

### Types of changes
- [x] Removes hard-coded  (non-breaking change enhances flexibility)

## Motivation and Context
Calling this script would set all VFATs to have `VThreshold1` of 100, which is problematic if a measurement is ongoing

## How Has This Been Tested?
It hasn't, but the change is trivial

### DIFF
--- a/gempython/tests/vfat_info_uhal.py
+++ b/gempython/tests/vfat_info_uhal.py
@@ -203,7 +203,7 @@ print
 print "--===== Last BX Latency Counter =========--"
 print
 counts = []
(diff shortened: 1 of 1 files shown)


bdorney:
Also in favor of just removing this line.

jsturdy:
I only commented it because having this type of operation is a fundamental part of testing communication, and envision that a future flag will be used for such an operation.
The current implementation is effectively a placeholder

mexanick:
Or add a comment why it is only commented.

Issue #146, opened 2017-09-14
How to handle masked broadcast read transaction results


jsturdy:
## Brief summary of issue
If a user performs a `broadcastRead` using a `mask`, the result object contains *only* entries corresponding to the mask, in increasing order, followed by `0`'s filling out the 24 entries

### Types of issue
- [x] Curiosity
- [x] Feature discussion (request comments for change in functionality)

## Current Behavior
```bash
[sturdy@gem904daq01 gemdev/vfatqc-python-scripts]% vfat_info_uhal.py -s2 -g0
14 Sep 2017 10:14:34.643 [7f0495621700] INFO - optohybrid_user_functions_uhal::getOHObject <> - gem.shelf01.amc02.optohybrid00: Success!
14 Sep 2017 10:14:34.643 [7f0495621700] INFO - optohybrid_user_functions_uhal::getOHObject <> - gem.shelf01.amc02.optohybrid00: Success!

--=======================================--

Firmware:     Version        Date
AMC     :       1.9.4   23/5/2017
OH      :    2.2.d.fb  05/05/2017
# mask is0xff000000
GEB0 SlotID::     0      1      2      3      4      5      6      7      8      9     10     11     12     13     14     15     16     17     18     19     20     21     22     23
     ChipID::0xf69f 0xff24 0xfa28 0xff60 0xfa40 0xfa17 0xf75b 0xff2c 0xfeeb 0xf99b 0xffe7 0xff4c 0xf747 0xff87 0xff40 0xff6b 0xfa20 0xffcb 0xff48 0xfa2c 0xff9b 0xffdc 0xff83 0xf6bc
   ContReg0::  0x36   0x36   0x36   0x36   0x02   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36   0x36
   ContReg1::  0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00   0x00
   ContReg2::  0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30   0x30
[...]

PR #148, opened 2017-09-20
Handling of TCDS and RCMS

jsturdy:
Some necessary updates for proper running with RCMS

## Description
* 4479731        (HEAD, origin/running-with-tcds, running-with-tcds) * placeholders for future functionality * ensure that some error messages are always printed the first time
* f8627b2        Boring
* 9fa3773        Improve some logging in the `GEMSOAPToolBox` code
* 7c3d2a2        * Modified printing of chip info function, registers in hex, DACs and settings in dec
* 2b74626        * Cleaned up some duplicate code
* 9a10413        Fixes related to sending parameters from RCMS * Udates to fix ParameterQuery bug * Obtain application state via SOAP for all managed applications, rather than via infospace lookup
* 4b2f4f6        Check whether we're trying to tell RCMS about the state change
* f0402fb        Try to make sure RCMS knows about the error
* 78cad64        Try to fail the FSM rather than raise exception
* ae07587        Raise exception to trigger a break out of the while loop
* 7aa6ff0        add FSM fail when composite state becomes 'Failed'
* 931db57        Added check for failed state in the state string checker
* 63522d9        Adding some debugging to the global state calculator to find where the update is getting stuck
* ed8a3c5        Update makefiles trying to make build process more uniform
* 54f5803        changes for FM running with tcds
* cd5b1e2        Patching in RCMS state information
*   b49f643      Merge branch 'running-with-tcds' of github.com:jsturdy/cmsgemos into running-with-tcds after rebasing
* 25dda7e      added TCDS update only if supervisor handles TCDS
* 87b11a2      added improved display of tables
* b513709      cleaned up some of the javascript for updating
* 71036a0      added more debugging for crashing TTS state handling
* 11b7f1d      changed some INFO to DEBUG
* 5e6f303      added debugging for TTS parsing
* 1848e37      added comments about some potential inefficiencies
* 749e41b      trying to handle TTS states
* 0092b2d      added TCDS update only if supervisor handles TCDS
* 5fdf78b      added improved display of tables
* b88b0fb      cleaned up some of the javascript for updating
* 7ae4e34      added more debugging for crashing TTS state handling
* 32cbbb8      changed some INFO to DEBUG
* 09587fd      added debugging for TTS parsing
* eebcc76      added comments about some potential inefficiencies
* 10b7eef      trying to handle TTS states

### Types of changes

### DIFF
--- a/gemhardware/src/common/amc13/AMC13Manager.cc
+++ b/gemhardware/src/common/amc13/AMC13Manager.cc
@@ -783,5 +783,3 @@ void gem::hw::amc13::AMC13Manager::endScanPoint()
                                                 p_appContext,p_appDescriptor,//getApplicationContext(),this->getApplicationDescriptor(),
                                                 const_cast<xdaq::ApplicationDescriptor*>(p_appZone->getApplicationDescriptor("gem::supervisor::GEMSupervisor", 0)));  // this should not be hard coded
 }
-
-
--- a/gemhardware/src/common/amc13/AMC13Manager.cc
+++ b/gemhardware/src/common/amc13/AMC13Manager.cc
@@ -468,8 +468,12 @@ void gem::hw::amc13::AMC13Manager::pauseAction()
   // what does pause mean here?
   // if local triggers are enabled, do we have a separate trigger application?
   // we can just disable them here maybe?
-  if (m_scanType.value_ == 2 || m_scanType.value_ == 3)
-    INFO("AMC13Manager::pauseAction disabling triggers for scan");
+  if (m_scanType.value_ == 2 || m_scanType.value_ == 3) {
+    INFO("AMC13Manager::pauseAction disabling triggers for scan, triggers seen this point = "
+	 << p_amc13->read(::amc13::AMC13::T1,"STATUS.GENERAL.L1A_COUNT_LO") - m_updatedL1ACount);
+    m_updatedL1ACount = p_amc13->read(::amc13::AMC13::T1,"STATUS.GENERAL.L1A_COUNT_LO");
+    INFO("AMC13Manager::timeExpried, total triggers seen = " << m_updatedL1ACount);
+  }
 
   if (m_enableLocalL1A) {
     if (m_enableLEMO)
@@ -781,13 +785,18 @@ void gem::hw::amc13::AMC13Manager::timeExpired(toolbox::task::TimerEvent& event)
 	p_amc13->stopContinuousL1A();
       }
     } else {
+      /*
+      // HACK
+      // when using external triggers, they should be stopped upstream of the AMC13 with a stop
+      // how do I hate thee? let me count the ways...
       p_amc13->configureLocalL1A(true, m_L1Amode, m_L1Aburst, m_internalPeriodicPeriod, m_L1Arules);
       p_amc13->enableLocalL1A(true);  // is this fine to switch to local L1A as a way to fake turning off upstream?
+      */
     }
     INFO("AMC13Manager::timeExpried, triggers seen this point = "
 	 << p_amc13->read(::amc13::AMC13::T1,"STATUS.GENERAL.L1A_COUNT_LO") - m_updatedL1ACount);
-    m_updatedL1ACount = p_amc13->read(::amc13::AMC13::T1,"STATUS.GENERAL.L1A_COUNT_LO");
-    INFO("AMC13Manager::timeExpried, total triggers seen = " << m_updatedL1ACount);
+    // m_updatedL1ACount = p_amc13->read(::amc13::AMC13::T1,"STATUS.GENERAL.L1A_COUNT_LO");
+    // INFO("AMC13Manager::timeExpried, total triggers seen = " << m_updatedL1ACount);
     endScanPoint();
   }
 }
--- a/gemhardware/src/common/amc13/AMC13Manager.cc
+++ b/gemhardware/src/common/amc13/AMC13Manager.cc
@@ -332,7 +332,8 @@ void gem::hw::amc13::AMC13Manager::initializeAction()
 
   // Use local trigger generator if config says to
   p_amc13->configureLocalL1A(m_enableLocalL1A, m_L1Amode, m_L1Aburst, m_internalPeriodicPeriod, m_L1Arules);
-  p_amc13->enableLocalL1A(m_enableLocalL1A);
+  // probably shouldn't enable until we're running? CHECKME
+  // p_amc13->enableLocalL1A(m_enableLocalL1A);
   p_amc13->write(::amc13::AMC13::T1,"CONF.TTC.T3_TRIG",0);
 
   // need to ensure that all BGO channels are disabled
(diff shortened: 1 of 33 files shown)


gemhardware/src/common/amc13/AMC13Manager.cc
```diff
@@ -307,22 +307,26 @@ void gem::hw::amc13::AMC13Manager::initializeAction()
...
   // enable SFP outputs based on mask configuration
   p_amc13->sfpOutputEnable(m_sfpMask);
-  /*
+
   // ignore AMC tts state per mask
-  if (m_ignoreAmcTts) {
-    p_amc13->write(amc13::AMC13Simple::T1, "CONF.AMC.TTS_DISABLE_MASK", 0xfff);
+  INFO("AMC13Manager::initializeAction m_amcIgnoreTTSList " << m_amcIgnoreTTSList);
+  m_ignoreAMCTTS = p_amc13->parseInputEnableList(m_amcIgnoreTTSList,true);
+  INFO("AMC13Manager::initializeAction m_amcIgnoreTTSList " << m_amcIgnoreTTSList
```
mexanick: Do we really want these two printouts as INFO? I would remove the line 320

gemhardware/src/common/amc13/AMC13Manager.cc
```diff
@@ -307,22 +307,26 @@ void gem::hw::amc13::AMC13Manager::initializeAction()
...
+       << " parsed as m_ignoreAMCTTS: " << std::hex << m_ignoreAMCTTS << std::dec);
+  if (m_ignoreAMCTTS) {
```
mexanick: what's the difference between `m_amcIgnoreTTSList` and `m_amcIgnoreTTS`? Do we really need both?

gemhardware/src/common/glib/GLIBManager.cc
```diff
@@ -162,6 +162,7 @@ void gem::hw::glib::GLIBManager::initializeAction()
   throw (gem::hw::glib::exception::Exception)
 {
   DEBUG("GLIBManager::initializeAction begin");
+  // FIXME make me more streamlined
```
mexanick: I suggest to use range-based loop here and later for alike places

gemhardware/src/common/glib/GLIBMonitor.cc
```diff
@@ -313,12 +384,201 @@ void gem::hw::glib::GLIBMonitor::buildMonitorPage(xgi::Output* out)
...
+             << "desc"
+             << "</td>"   << std::endl
+             << "</tr>"   << std::endl;
+      }
+      break;
+    }
+  }
+  *out << "</tbody>" << std::endl
+       << "</table>" << std::endl
+       << "</div>"   << std::endl;  // closes Trigger Status tab
```
mexanick: Generic comment: this is all referring to the GLIB while being common to any kind of AMC board. Shouldn't we start moving/renaming the stuff as we are already having a large update?

gempython/tools/optohybrid_user_functions_uhal.py
```diff
@@ -171,7 +171,9 @@ def broadcastRead(device,gtx,register,mask=0xff000000,debug=False):
         time.sleep(0.1)
         pass
 
-    return readBlock(device,"%s.Results"%(baseNode),24)
+    # bitcount = bits not set in mask
+    bitcount = 24
```
mexanick: why it is hard-coded here? If it is a temp plug before we implement proper masking, shouldn't it be placed where we are going to get the mask?

gemhardware/src/common/amc13/AMC13Manager.cc
jsturdy: Good catch, I think the second should be `m_ignoreAMCTTS`

gemsupervisor/src/common/GEMSupervisor.cc
```diff
@@ -270,18 +285,23 @@ void gem::supervisor::GEMSupervisor::initializeAction()
...
-        if (((*j)->getClassName()).rfind("tcds::") != std::string::npos) {
-          INFO("GEMSupervisor::initializeAction Halting " << (*j)->getClassName()
-               << " in case it is not in 'Halted'");
-          // need to ensure leases are properly respected
-          gem::utils::soap::GEMSOAPToolBox::sendCommand("Halt", p_appContext, p_appDescriptor, *j);
-        } else {
-          INFO("GEMSupervisor::initializeAction Initializing " << (*j)->getClassName());
-          gem::utils::soap::GEMSOAPToolBox::sendCommand("Initialize", p_appContext, p_appDescriptor, *j);
+      // if (!m_gemRCMSNotifier.getFoundRcmsStateListenerParameter()) {
+      if (true) {
```
mexanick: ???

gemhardware/src/common/amc13/AMC13Manager.cc
jsturdy: The `m_amcIgnoreTTSList` is the `std::string` setting that lists which `TTS` inputs should be ignored on the backplane, and the `m_amcIgnoreTTS` is a `boolean` to check if this needs to happen.
However, I see now that it's not done correctly in the setting step, so I will change `m_ignoreAMCTTS` from `boolean` to `uint32_t` and write that to the `ttsDisableMask` function

gemhardware/src/common/glib/GLIBMonitor.cc
jsturdy: That's quite a bit more of an update, so I'll do that once the content is stabilized.

gempython/tools/optohybrid_user_functions_uhal.py
jsturdy: As you say, it is a placeholder to have me remember to actually implement it (if we decide to do so in #146)

gemutils/src/common/soap/GEMSOAPToolBox.cc
```diff
@@ -306,11 +315,16 @@ bool gem::utils::soap::GEMSOAPToolBox::sendCommandWithParameter(std::string cons
     xoap::SOAPBodyElement paramValue = env.getBody().addBodyElement(soapcmd);
     paramValue.addTextNode(toolbox::toString("%d", parameter));
 
-    // std::string tool;
+    std::string tool;
     // xoap::dumpTree(msg->getSOAPPart().getEnvelope().getDOMNode(),tool);
+    msg->writeTo(tool);
+    INFO("GEMSOAPToolBox::sendCommandWithParameter SOAP msg " << tool);
+    tool.clear();
```
mexanick: This is a repeatedly used construction and could be streamlined as a function

gemsupervisor/src/common/GEMSupervisor.cc
jsturdy: It may at some point be necessary/useful to check if RCMS is controlling the supervisor, but this turned out to not be mandatory to a particular problem we were addressing in the `rcms` <--> `xdaq` cross communication, so I switched it to this so that everything would still "work", but that I could easily get back and properly implement the feature that I was trying to flesh out

gemutils/src/common/soap/GEMSOAPToolBox.cc
jsturdy: Indeed, #151 created to address this

gemhardware/src/common/glib/GLIBManager.cc
jsturdy: It's not quite as simple as that, as I may have mentioned before.
The `SOAP` commands and the `parameterGet`/`parameterSet` commands are expecting the populated object at specific indices in the `SOAP` message
So as has been my hope/plan for a while, this will require a bit of careful thought to make sure it gets handled properly.

gemhardware/src/common/amc13/AMC13Manager.cc
jsturdy: New implementation should now be pushed

jsturdy:
0532654 adds support for #160

jsturdy:
Need a few tweaks, and then will make a new release, version 1.0.0

PR #156, opened 2017-10-10
Update amc_user_functions_uhal.py

jsturdy:
Add print functions that I want the DOC to be able to obviously put into an elog without thinking

## Description

Added two functions to print the `TTC` and `SCA` status blocks

### Types of changes


jsturdy:
ready for review

PR #168, opened 2018-01-31
Bring in travis and gitlab CI functionality

jsturdy:
## Description
Brings changes showing preliminary CI examples with both `gitlab-ci` and `TravisCI` (using the same `docker` container that is generated for the `gitlab-ci` build.

### Types of changes

## Motivation and Context
This PR paves the way for enhanced maintainability, and will provide (with some additional work, already started in other branches) the ability to have `github` automatically create releases (we can select when and how this happens)

### How Has This Been Tested?
The `travis` builds have been tested, and other code hasn't been touched (specifically for this purpose)
* Builds are tested against both `slc6` and `cc7`

### Additional notes:
Unfortunately, I don't have a good way to currently do a pre-merge hook to change some text that is `jsturdy` specific...  I investigated a couple of options, but nothing good came up, so there will be a few manual changes from `jsturdy` to `cms-gem-daq-project` necessary


mexanick:
Giving the huge PR another look, I believe it should not go to the generic-amc-release-v3, but rather requires a new branch. After it will pass the testing, it can be merged quiet easily with all the release  branches as it doesn't really interfere with XDAQ-related developments

jsturdy:
Actually, its base is `generic-amc-release-v3`, so it doesn't really belong anywhere else, and will become part of all future branches when `generic-amc-release-v3` migrates to `master`/`release-v3`, with new tag and maintenance support for backports of `v2b` specific features

PR #171, opened 2018-02-19
Changed sturdy/jsturdy specifics to cms-gem-daq-project

jsturdy:
Removing `sturdy` and `jsturdy` references, changing to `cms-gem-daq-project`


jsturdy:
Not able to find comparable SVG logos for all integrations (also, not all integrations currently activated), so leaving table header as simple text

PR #184, opened 2018-06-08
New Feature: Adding v3 XHAL gempython functionality

bdorney:
## Description

Per our discussion in today's ops meeting.  These should be the only files necessary for working with the v3 hardware.

### Types of changes

## Motivation and Context

By placing functionality here we reduce maintenance headache and complexity.

## How Has This Been Tested?

Previously tested before being committed to `generic-amc-release-v3-RPC-Playground`

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -5,26 +5,18 @@ from gempython.utils.gemlogger import colors,colormsg
 from gempython.utils.nesteddict import nesteddict
 from time import sleep
 
-from reg_utils.reg_interface.common.reg_xml_parser import Node, parseXML, parseInt
+from reg_utils.reg_interface.common.reg_base_ops import rpc_connect, rReg, writeReg
+from reg_utils.reg_interface.common.reg_xml_parser import getNode, Node, parseInt, parseXML
+
+from xhal.reg_interface_gem.core.reg_extra_ops import rBlock
 
 import logging
 
 gMAX_RETRIES = 5
 gRetries = 5
 
-class ctp7Params:
-    # Key (shelf,slot); val = eagleXX
-    # See: https://twiki.cern.ch/twiki/bin/view/CMS/GEMDAQExpert#List_of_CTP7_s
-    cardLocation = { 
-            (3,2):"eagle26", #QC8
-            (2,5):"eagle60",
-            #(?,?):"eagle61",
-            (1,3):"eagle33", #P5
-            (1,2):"eagle34"} #Coffin
-            #(1,?):"eagle64"}
-
 class HwAMC:
-    def __init__(self, slot, shelf=1, debug=False):
+    def __init__(self, cardName, debug=False):
         """
         Initialize the HW board an open an RPC connection
         """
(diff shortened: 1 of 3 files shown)


gempython/tools/amc_user_functions_xhal.py
```diff
@@ -0,0 +1,343 @@
+import os, signal, sys
+sys.path.append('${GEM_PYTHON_PATH}')
+
+from ctypes import *
+from gempython.utils.gemlogger import colormsg
+from gempython.utils.nesteddict import nesteddict
+from gempython.utils.wrappers import envCheck
+from time import sleep
+
+from rw_reg import Node, parseInt, parseXML
```
jsturdy: `rw_reg` is `xhal`/`reg_utils` now right?
suspect that as is, this will not work

gempython/tools/amc_user_functions_xhal.py
```diff
@@ -0,0 +1,343 @@
+import os, signal, sys
+sys.path.append('${GEM_PYTHON_PATH}')
```
jsturdy: this is not compatible

gempython/tools/amc_user_functions_xhal.py
```diff
@@ -0,0 +1,343 @@
...
+from time import sleep
+
+from rw_reg import Node, parseInt, parseXML
+
+import logging
+
+gMAX_RETRIES = 5
+gRetries = 5
+
+class colors:
```
jsturdy: this is defined in `utils/gemlogging.py` or somewhere in there

gempython/tools/optohybrid_user_functions_xhal.py
```diff
@@ -0,0 +1,259 @@
+import os, signal, sys, time
+sys.path.append('${GEM_PYTHON_PATH}')
```
jsturdy: incompatible

gempython/tools/amc_user_functions_xhal.py
mexanick: atm it is in xhal. But won't exist anymore after integration of reg_utils. Will be in:
`reg_utils.reg_interface.common.reg_xml_paser`
However it is fine to go as is right now and change later with inclusion of revised xhal/reg_utils

gempython/tools/amc_user_functions_xhal.py
jsturdy: though i see it currently exists in other places and just needs to be removed...

gempython/tools/amc_user_functions_xhal.py
bdorney: yes at the moment all that is needed is to setup the `xhal` env.

When will integration of `reg_utils` be performed?  If it is a short time scale I can make this change now.  Otherwise I would prefer to leave it *as is* for the time being so a working release could be generated.

gempython/tools/amc_user_functions_xhal.py
bdorney: Okay, so I understand that all instances of:

```
sys.path.append('${GEM_PYTHON_PATH}')
```

Should be removed.  Correct?

gempython/tools/amc_user_functions_xhal.py
bdorney: Okay, I'll remove it and change to an import.

gempython/tools/amc_user_functions_xhal.py
jsturdy: if by "fine", you mean it won't currently work, i agree (dependence on removed `GEM_PYTHON_PATH`)

mexanick: `reg_utils` shoud be integrated next week. Actually, it is ready now barring the documentation completion in travis/docker what I fight right now.

jsturdy: better to decouple deployment from documentation

gempython/tools/amc_user_functions_xhal.py
jsturdy: roger, but i suppose this will render this current code inoperable, as per `rw_reg` business...

gempython/tools/amc_user_functions_xhal.py
mexanick: this will be also in reg_utils... or we might need some moar refactoring

gempython/tools/amc_user_functions_xhal.py
mexanick: well, documentation is a sort of integration test here...

bdorney: > if by "fine", you mean it won't currently work, i agree (dependence on removed GEM_PYTHON_PATH)

I've tested this in cmsgemos `v0.3.1` it works fine...I don't follow.

jsturdy: because while you test, `PYTHONPATH` was explicitly set to include the location of the `rw_reg.py` file

gempython/tools/amc_user_functions_xhal.py
jsturdy: regarding this PR, this is probably the only comment that *must* be implemented, as the others will be future cleanup

bdorney:
Cross-checking API now; think I'm running into env setup issues.  Will update.

bdorney:
Please review at earliest convenience.

gempython/tools/optohybrid_user_functions_xhal.py
```diff
@@ -0,0 +1,248 @@
...
+        if self.parentAMC.fwVersion < 3:
+            if scanReg not in self.KnownV2bElScanRegs:
+                print("Parent AMC Major FW Version: %i"%(self.parentAMC.fwVersion))
+                print("V2b Electronics Behavior Detected")
+                print("scanReg %s Not Known")
+                print("Available scanReg values are:", self.KnownV2bElScanRegs)
+                sys.exit(os.EX_USAGE)
+        else:
+            if "CFG_" in scanReg:
+                scanReg = str.replace("CFG_","")
```
mexanick: Good catch by codacity: it is a bug! Should be 
```python
scanReg=str.replace(scanReg,"CFG_","")
```

PR #204, opened 2018-09-24
cleanup hardcode hacks

jsturdy:
## Description
* Remove several hard-coded hacks

### Types of changes

## Motivation and Context
Makes code more flexible by default

## How Has This Been Tested?
Operating at P5

* When this is rebase+merged I'll put it on `master`, and create the `1.0.0` tag
 * Only a few P5 operation specific code will then go into this tag series
* Then will create the new branch for our `2.0.0` release line
 *  prerel tag will probably be `1.99.0`
 * Code for clean and proper `xhal` inclusion will be going into this branch

### DIFF
--- a/gemhardware/include/gem/hw/glib/GLIBManager.h
+++ b/gemhardware/include/gem/hw/glib/GLIBManager.h
@@ -87,6 +87,7 @@ namespace gem {
             xdata::Integer crateID;
             xdata::Integer slotID;
             xdata::String  cardName;
+            xdata::String  birdName;
 
             //configuration parameters
             xdata::String controlHubAddress;
@@ -110,6 +111,7 @@ namespace gem {
                  << "crateID:"  << crateID.toString()  << std::endl
                  << "slotID:"   << slotID.toString()   << std::endl
                  << "cardName:" << cardName.toString() << std::endl
+                 << "birdName:" << birdName.toString() << std::endl
 
                  << "controlHubAddress:" << controlHubAddress.toString() << std::endl
                  << "deviceIPAddress:"   << deviceIPAddress.toString()   << std::endl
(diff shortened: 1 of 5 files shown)


gemhardware/include/gem/hw/glib/GLIBManager.h
```diff
@@ -87,6 +87,7 @@ namespace gem {
             xdata::Integer crateID;
             xdata::Integer slotID;
             xdata::String  cardName;
+            xdata::String  birdName;
```
bdorney: what is the difference between `birdName` and `cardName`?

bdorney: I see thanks

bdorney: Maybe we could add a commented line here to illustrate these distinctions?

gemsupervisor/src/common/GEMSupervisor.cc
```diff
@@ -384,8 +384,6 @@ void gem::supervisor::GEMSupervisor::configureAction()
     m_globalState.update();
   }
 
-  pausePhaseMonitoring();
```
mexanick: What was the reason to pause phase monitoring an was this change tested?

jsturdy: The `pause`/`resume` was implemented when we didn't have the `memhub` protection against potentially colliding transactions.

jsturdy:
Actually, based on the problem of today, I want to add three more things to this PR, though they can probably wait for the "maintenance" branch:
* SCA `HardReset` commands
* VFAT watchdog while running
* monitoring of links, and removing from the active links list, links which have lost communication
 * if the SCA `HardReset` works, this could be attempted, and a subsequent reconfiguration
 * if the SCA `HardReset` fails, this link would be excluded from any ongoing configuration (if the config was destroyed, this information would be reset)
* possibly also trigger `RunningDegraded` if the work is not too much for both the link lost issue, as well as any link `autokilled`

Issue #197, opened 2018-08-17
Provide DAQ monitoring interface and integrate RPC calls to retrieve monitorables


mexanick:
## Brief summary of issue

- Implement firing of corresponding items to the infoSpace
- Implement web templates
### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

mexanick:
thanks for the detailed comment! My thoughts are inline:

> * access to a given hardware endpoint (AMC) should be parallelized, as this will be the level at which no more parallelization can take place, all register transactions will be serilaized

The plan is to have an AMCDevice or AMC manager class and provide one instance of this class per AMC board. Each object will be independent and can leave in its own thread (and thread behavior can be additionally monitored by the means of XMAS). This object will manage a connection to the given AMC board and provide all the communication to the RPC front-end.

> * monitoring should be able to be switched on/off without breaking things

This would be re-implemented from the base GEMMonitor class. The RPC connection itself can stay active, but we can stop sending update requests

>  * maybe even having the monitoring timers have their timeouts changed on the fly

Yes, we will try to implement exactly this 

* endpoints should be able to be masked easily

This will require some rearrangements in the code, but in principle is doable, no problems

> * endpoints that are not reachable should not waste resources, even if they're configured to be available
>   * report "link dead" and don't try to obtain all monitorables
>   * could be a warning/error, especially in the case where the disconnected link is not expected, but if performing a FW update/reboot, this would be a perfectly acceptable interruption

Link access errors can be intercepted. So far they are just ignored (well, will be shown as errors in the interface, but the attempts to pull out information will be happening. In this regards we are wasting the resources, but as soon as link comes on we have it in the monitoring again
 
>  * could be nice to automatically include endpoints that become available without a restart of core software, but this is definitely not a deal breaker

We can foresee few ways of this... Either through a call from a HwManager which performed the recovery (kinda "notify the monitor we have recovered something" or through a general connectivity polling (some resources will be wasted, but not that much)

>  * need to understand under which circumstances the RPC connection needs to be reinitialized?

The RPC connection error raises a corresponding exception which can be intercepted. The follow up procedure has to be defined... Either simple automated reconnection, or a call to certain recovery procedure... TBD

>   * with `uhal` if the hardware becomes available again, things will automatically start working, with RPC we probably have to reopen the connection on the SW side (possibly a task for #196)

[...]

Issue #196, opened 2018-08-17
Integration of xHAL RPC manager with the cmsgemos


mexanick:
## Brief summary of issue

Need to integrate the xHAL library into HwManager base class somehow. Please let me know of your thoughts. So far I was thinking about providing an RPCManager pointer to the HwDevice parallel to the uhal::Device
### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

PR #51, opened 2016-11-19
Update AMC13Manager.cc

jsturdy:
AMC13 code base changed upstream to fix leak
API change requires some changes in BGO setup

PR #205, opened 2018-10-02
Develop daq monitoring

mexanick:
## Description

Adding TTC and Optical links tables, almost everything is there at the moment. Adding CSS for representation as well. 
Known issues: the web interface should not be called until all the DaqMonitor objects initialize
Still need more testing to check the actual functionality, for the moment only the proof of concept is received due to address table mismatch at the integration stand. Also javascript is not attached, so the automated update doesn't happen.
### Types of changes

Addressing #196 and #197  
## How Has This Been Tested?

Tested with dummy system at integration stand @904
### Screenshots (if appropriate):
screen shot 2018-10-02 at 1 40 00 pm

## UPDATE
The bugs related to retrieving correctly OH info were fixed
screen shot 2018-10-02 at 3 29 37 pm

## Checklist:


mexanick:
ok, all bugs fixed, the monitoring works including javascript part. Ready to review. 
All requests about adding more monitorables should go into issues

bdorney:
As a general comment, either it's a low quality image, but I cannot read the numbers at all from the screen shot of the xdaq web image.

gemdaqmonitor/src/common/DaqMonitor.cc
```diff
@@ -253,6 +196,7 @@ void gem::daqmon::DaqMonitor::updateDAQOHmain()
 {
   DEBUG("DaqMonitor: Update DAQ OH main table");
   req = wisc::RPCMsg("amc.getmonDAQOHmain");
+  req.set_word("NOH",NOH);
```
bdorney: I think @jsturdy raised this point, rather than setting an NOH word in the RPC msg, shouldn't the RPC module just get NOH from the FW?  This is what I've asked Andrew to do in [ctp7_modules #51](https://github.com/cms-gem-daq-project/ctp7_modules/pull/51#pullrequestreview-161229087)

mexanick:
well, the screenshot quality isn't that great and it is taken from a large screen. Indeed, the font size is small (otherwise tables won't fit). However it is readable on 27" screen and if needed, the page design can be altered in future...

gemdaqmonitor/src/common/DaqMonitor.cc
mexanick: this might be updated later. Current version of RPC module is like that, so the `NOH` parameter is set. An update to daq module is planned anyway (first, this functionality will be in a different module, second, some of the functions will be altered)

PR #207, opened 2018-10-08
Providing additional monitoring data

mexanick:
## Description

Adding `Input Enable` and `autokill` masks and temperature and voltage monitoring over `SCA` and `Sysmon`. ADC calibration still has to be revisited and PT monitoring has to be cross-checked...
### Types of changes

## How Has This Been Tested?

Tested at integration stand @904
## Checklist:

### DIFF
--- a/gemdaqmonitor/include/gem/daqmon/DaqMonitor.h
+++ b/gemdaqmonitor/include/gem/daqmon/DaqMonitor.h
@@ -46,6 +46,17 @@ namespace gem {
       void updateTRIGGERmain();
       void updateTRIGGEROHmain();
       void updateOHmain();
+      void updateOHSCA();
+      void updateOHSysmon();
+      //all voltages are divided by 3 on the OH board by a voltage divider (temperature sensor voltage is not divided).
+      //All values are in units of ADC counts, and this is a 12bit ADC with a range of 0.0 - 1.0V, so each count is 1V / 4095 = 0.244mV.
+      double scaVconv(uint32_t val);
+      //There's not much information about the SCA chip temperature units, but some information can be found in the SCA manual page 52, which shows linear dependence of temperature from -37.5 deg C to 79 deg C corresponding to 790 counts for -37.5degC and going down to about 570 counts for 79degC.
+      double scaTconv(uint32_t val);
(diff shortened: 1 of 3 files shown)


bdorney:
Saw this last week, from v2b side everything looks good. Thanks for fast turn around. Now just need FW fix to SCA monitoring block @evka85

gemdaqmonitor/include/gem/daqmon/DaqMonitor.h
```diff
@@ -46,6 +46,17 @@ namespace gem {
...
+      double scaTconv(uint32_t val);
+      //For PT100 measurements, a more complex conversion is necessary to covert the ADC counts to temperature. The way it works is that the ADC supplies 100uA current that goes through the PT100 sensor to ground, and the voltage drop over the sensor then corresponds to the temperature. The exact part number of the PT100 sensor we are using is P0K1.1206.2P.B, which has a resistance of 100 Ohms at 0 deg C, and changes by 0.385% per each degree C cumulatively (more info can be found here: http://www.farnell.com/datasheets/2207165.pdf?_ga=2.247382788.1064362929.1535377707-998824428.1535377707 and here: https://www.intech.co.nz/products/temperature/typert/RTD-Pt100-Conversion.pdf). This means that at 0 deg C the voltage drop over the PT100 will be 100uA * 100 Ohms = 10mV, so the ADC will read 1.0V - 10mV = 0.99V, and if the temperature is 30 degC, the PT100 will have a resistance of 111.67 Ohms, and so the voltage drop will be 100uA * 111.67 Ohms = 11.167mV, so the ADC will read 1.0V - 11.167mV = 0.988833V.
+      double scaPTconv(uint32_t val);
+      double sysmonTconv(uint32_t val);
+      double sysmonVconv(uint32_t val);
```
jsturdy: In principle, for all of these conversions, I have a slight preference that they happen in some generic place, such that they are generally accessible, i.e., if the manager code needs to do this ever, or if some exported `python` code wants this, it would be better somehow as part of the `gem/hw` libraries

gemdaqmonitor/src/common/DaqMonitor.cc
```diff
@@ -19,8 +20,8 @@ gem::daqmon::DaqMonitor::DaqMonitor(const std::string& board_domain_name,log4cpl
 {
   DEBUG("DaqMonitor::DaqMonitor:: entering constructor");
   if (isConnected) { //TODO Add to the app monitoring space? Need to know in order to mask in web interface the boards which failed to connect
-    this->loadModule("amc", "amc v1.0.1");
-    DEBUG("DaqMonitor::DaqMonitor:: amc module loaded");
+    this->loadModule("daq_monitor", "daq_monitor v1.0.1");
```
jsturdy: how important is this version hardcoding, and can it be removed in favour of some better way, i.e., a compile time variable that can be driven by the make process (or even extracted from the linked library)?

gemdaqmonitor/include/gem/daqmon/DaqMonitor.h
bdorney: Indeed, python code could use this either in `monitorTemperatures.py` or in a script to analyse that data.

mexanick: It was requested that the monitoring app should be able to run as a standalone application, thus all relevant conversions has been made here. This also significantly simplifies integration of the monitoring package with the rest of the cmsgemos. I'm not much against of moving this to some other place, but I don't see any clear preference. Moreover, I would like to stress that this might very well be a temporary solution as the conversion factors aren't yet fully understood. Another question is whether we would like to store monitoring information in ADC counts, or in physical values. In latter case, all the conversion can happen in the monitoring app on-the-flight.

PR #208, opened 2018-10-10
Modify the spec file for owned files

jsturdy:
## Description
Packaging was resulting in conflicting ownership, this change makes ownership more explicit in each package

### Types of changes


mexanick:
Is it stale? @jsturdy

jsturdy:
Probably, if not, it will probably get resolved elsewhere

PR #210, opened 2018-10-26
Develop daq monitoring

mexanick:
Adding start/stopping actions and JSON file dumping
## Description

Now the monitoring can be manually paused and resumed if required. The JSON file with monitorables is dumped every approx 20 seconds. Some timers still require certain improvements, however this goes out of scope for the current PR and will be worked out later.
### Types of changes

## Motivation and Context

Pausing the monitoring is required when other more important register-intensive operations have to be executed, such as phase alignment. 
## How Has This Been Tested?

Tested at 904 and P5 setups
## Checklist:

### DIFF
--- a/gemdaqmonitor/Makefile
+++ b/gemdaqmonitor/Makefile
@@ -28,6 +28,7 @@ IncludeDirs+=/opt/xhal/include
 
 DependentLibraryDirs+=/opt/xhal/lib
 DependentLibraries+=xhal
+DependentLibraries+=boost_iostreams
 
 include $(XDAQ_ROOT)/config/Makefile.rules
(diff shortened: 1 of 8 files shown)


mexanick:
Right now there's no difference between "Stop" and "Pause" and the monitoring is started by default when you start the application - hence there's no "Start" button, but there's a "Resume". You may want to ask why then a "Stop" button exists? It could be useful in future to have "Stop" action to de-configure the monitoring in addition to stopping sending the "getMonitorables" kind of commands. 
Also it is foreseen that the actions will have attached listeners and an external monitoring managememnt from the FM/GEMSupervisor will be implemented at some point.

PR #216, opened 2018-11-07
[V3] Updating List of Possible DACs for DAC Scans

bdorney:
## Description

Added `CFG_BIAS_PRE_I_BSF` and `CFG_CAL_DAC` to list of possible DAC scans for V3.

### Types of changes

## Motivation and Context

Registers were previously omitted since we did not know the correspondence between monitor select and CTP7 address space.

## How Has This Been Tested?

Change was trivial.

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.

### DIFF
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -17,7 +17,7 @@ maxVfat3DACSize = {
         #1:???
         2:(0xff,"CFG_BIAS_PRE_I_BIT"),
         3:(0x3f,"CFG_BIAS_PRE_I_BLCC"),
-        #4:???
+        4:(0x3f,"CFG_BIAS_PRE_I_BSF"),
         5:(0xff,"CFG_BIAS_SH_I_BFCAS"),
         6:(0xff,"CFG_BIAS_SH_I_BDIFF"),
         7:(0xff,"CFG_BIAS_SD_I_BDIFF"),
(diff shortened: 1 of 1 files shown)


bdorney:
> is thought going into how to avoid this duplication long term?

From the CTP7 side I originally thought to incorporated this into the lmdb and then have a lookup be used.  But this was not desired by @mexanick.  So the strategy there was to have a hardcoded set of info:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/4d82a717400fd96cb6f18104b58b93a1e8751036/src/calibration_routines.cpp#L1370-L1401

Which perhaps it's not the best solution but since this is specific to the ASIC it could be done just here.

For resolving the duplication between cmsgemos and xhal this will be resolved when we expose things to boost.  Then some XHALInterface in cmsgemos (I don't know how this would fit into a HwDevice since this is a calibration routine...) would obtain an `std::vector`.  This would of course be expected to be a specific length as it is now but `xhal` would no longer need to know how long removing the need for:

https://github.com/cms-gem-daq-project/xhal/blob/a9afa48377eaf64538d6b6fe34070f3e2aaffe32/xhalcore/include/xhal/rpc/calibration_routines.h#L9-L40

To be explicitly clear, right now the duplication exists only because we are passing arrays to the DLL's from python to C++.  We can work only with `std::vector`'s in the future, expose them to python as a list, and then expose a single hardcoded struct/container (in a specific location) as a python `struct`.

To me this would avoid duplication.  We don't even need to hard code it if we store this info in the DB and just perform a DB lookup for the "dacSelect" of interest.

What do you think?

jsturdy:
OK, let's revisit that once we're further along that path, thanks!

Issue #224, opened 2018-12-03
Can't generate doc


lmoureaux:
## Brief summary of issue

I'm experiencing trouble when trying to add documentation. Namely,
```
make doc
```
generates the following output:
```
Generating doxygen
make: *** [doc] Error 1
```

Removing the appropriate `>/dev/null` from the Makefile, I get:
```
Generating doxygen
error: configuration file ./doc/cmsgemos.cfg not found!
```

Indeed the file doesn't exist. Am I doing something wrong?
### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

`make doc` works

## Current Behavior

See above

## Possible Solution (for bugs)

Include the missing file

## Context

Willing to document changes in PR #222

lmoureaux:
@jsturdy (only admins can assign)

jsturdy:
I'm pretty sure (read, certain) that `doc` target is not validated.
In my local area, I had put the configs in `docs/`, but have not yet pushed the config.
I will can run through and make sure things build with the given config and add a patch that you can `rebase` off.

jsturdy:
You should be able to `fetch` and `rebase` from `cms-gem-daq-project:feature/gemonlinedb` to fix this in your working tree.  If any issues remain reopen this.

lmoureaux:
```
error: tag OUTPUT_DIRECTORY: Output directory `/afs/cern.ch/user/l/lmoureau/GEM/gemonlinedb/cmsgemos/docs/build/' does not exist and cannot be created
```
Works if I create the directory by hand. Only maintainers can reopen issues in this repo.

jsturdy:
OK, `build` doesn't have anything persistent in it, so #227 should fix that.
There *shouldn't* be anything else preventing the doc building, but doc building is really my least concern at present...

lpetre-ulb:
The documentation generation is still failing with the update build system. On the first attempt, one gets the following error:
```bash
$ scl enable devtoolset-8 -- make doc
[...]
Generating doxygen
Warning: Tag `SYMBOL_CACHE_SIZE' at line 304 of file ./doc/cmsgemos.cfg has become obsolete.
To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
Warning: Tag `SHOW_DIRECTORIES' at line 498 of file ./doc/cmsgemos.cfg has become obsolete.
[...]

PR #222, opened 2018-11-28
Online DB configuration objects

lmoureaux:
## Description

This PR is to track my progress on DB configuration objects (#213).

Currently implemented:
* Python script to generate `xsd` templates and class definitions from a list of fields.
* Integration of the script into the `Makefile` for the VFAT3 chip configuration.
* XMLSerializationData class representing the structure of XML documents in C++
* Function to turn XMLSerializationData into a DOM tree
* Function that fills an XMLSerializationData object with the contents of a DOM tree
* Tests of serialization functionality:
     * Create a DOM tree and check that it isn't empty
     * Create a DOM tree and validate it against the schema
     * Create a DOM tree, load it and check that we get back the initial object

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

#213, #214 and #215 for VFAT3 chip and channel configuration data.

## How Has This Been Tested?

* Checked that a very basic code importing the generated class compiles
* The `xml` file in the repo can be checked using the generated `xsd` template
* Tests compile and yield the expected results
* The tests have been run under Valgrind and no leaks are found 

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly. [Link to documentation](http://homepage.iihe.ac.be/~lmoureau/GEM/cmsgemos/doc/build/html/md_gemonlinedb_gemonlinedb.html)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #225, opened 2018-12-03
doxygen hotfix

jsturdy:
* Add doxygen config file
* Update `Makefile` for `doc` target


jsturdy:
For #224 (related #222)

PR #227, opened 2018-12-04
Update Makefile for building docs

jsturdy:
Create `doc/build`, required by `doc` `make` target
As per #224

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,8 @@ dbgprofile: $(SUBPACKAGES)
 
 doc:  $(SUBPACKAGES)
 	@echo "Generating doxygen"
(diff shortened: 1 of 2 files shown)


lmoureaux:
Thanks for the hotfix. The required directory is `docs/build` (not `doc/build`) as set at https://github.com/cms-gem-daq-project/cmsgemos/blob/9a423631a9b42a1d6d87bbc0e3461f8ed9fdfb9b/doc/cmsgemos.cfg#L45
It's not urgent since creating the missing directory is easy.

jsturdy:
Ah, I see, I didn't check the cfg file, only the `Makefile`

PR #232, opened 2019-01-24
New Feature [V3 ]: Python Tool For Setting Latency Parameters compatible w/XDAQ

bdorney:
## Description

I've updated the `scan_latency.py` tool we've used at P5 for v3 software & hardware. 

### Types of changes

## Motivation and Context

Needed something to use when scanning latency both for several measurements (e.g. cosmics, cross-talk, etc...) 

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1078467

## Checklist:


mexanick:
This tool is only setting the parameters for the latency scan, but does not take the scan *per se*, does it? If this is the case, I would rename it. Also, I would rename the `main` function and make it easily accessible for the external code (e.g. call it something like `set_scan_params` or so)

bdorney:
Yes I agree and I've done some renaming based on the provided feedback.

Issue #235, opened 2019-02-07
[feature] Populate configuration objects from TStore result


jsturdy:
## Brief summary of issue
The `TStore` library returns a table result of the DB query, and we need to fill the configuration objects from #231 with the resulting information

### Types of issue
- [x] Feature request (request for change which adds functionality)

lpetre-ulb:
It is unlikely that an Oracle SQL database is going to be used (and by extension the `TStore` library). Discussion about the database system to be used is in [progress on GitLab](https://gitlab.cern.ch/cmsgemonline/cmsgemos/-/issues/68). New issues will be opened with the outcome of this discussion.

PR #237, opened 2019-02-07
Populate configuration objects from TStore result

lmoureaux:
## Description

This adds a static function template `DBInterface::convertRow` that converts a `xdata::Table::Row` into a configuration object and tests that pull data from the database.

### Types of changes

## Motivation and Context

Requested in #235

## How Has This Been Tested?

Tests pass. See `tests/testTStoreSerialization.cc`.

## Checklist:


lmoureaux:
@mexanick Please review this one before the [WIP] #238 that was built on top of it ;-)

lmoureaux:
> Why should we have cases where a valid integer is not stored as `int` and must be stored as `string`?

I can't see a valid use case, but that's how the DB is constructed atm. I can change the software if the DB is updated, but I need to support the current layout somehow.

jsturdy:
Merging blocked by #238 for sequential (better in future to branch from same parent, and rebase once other features have merged, unless strictly dependent on everything in a separate feature PR)

lmoureaux:
Here's how TStore chooses the type based on parameters of the `NUMBER` type `precision` and `scale` [1]:
```C++
if (scale==0) { //if it's an integer
    if (precision<std::log10((long double)INT_MAX)) return "int";
         return "string";
    } else {
        if (precision<=FLT_DIG) return "float";
        if (precision<=DBL_DIG) return "double";
            return "string";
     }
}
```

Most registers are stored in the database are with `precision=10` and `scale=0`. This corresponds to numbers with at most 10 digits, covering a range slightly larger than 32 bits can offer. This means that the expression `precision<std::log10((long double)INT_MAX)` evaluates as `false` and the type deduced by `TStore` is `string`.

I definitely need to update `std::stoi` -> `std::stoul` and add bounds checks since I store values as `std::uint32_t`, but I can't do much more.

[1] `daq-tstoreutils-3.1.2-1.cmsos14.centos7.gcc4_8_5.src.rpm`: `/usr/local/src/xdaq/baseline14/daq/tstore/utils/src/common/OracleConnection.cc` lines 487+

jsturdy:
> I definitely need to update `std::stoi` -> `std::stoul` and add bounds checks since I store values as `std::uint32_t`, but I can't do much more.

Every setting that we're pulling out of the DB will be a register value, so at most a 32-bit unsigned.
In some cases they'll be 8- or 16-bit, but I don't know if there's a strong reason to specialize that further than the 32-bit in the `TStore` conversion step...

lmoureaux:
> In some cases they'll be 8- or 16-bit

These are the only cases where the `TStore` conversion is actually useful.

> I don't know if there's a strong reason to specialize that further than the 32-bit in the `TStore` conversion step

1. Because the `TStore` conversion is always 32 bits signed, it cannot handle all possible unsigned values on 32 bits.
2. Since DB types don't map directly to C++ types, the `TStore` interface isn't the most adapted and fails to hide some implementation details. The conversion from `string` is a workaround. More details [in the code](https://github.com/cms-gem-daq-project/cmsgemos/blob/bf681b0090b83fd9d299c60bb4d0ec1ac2157b9c/gemonlinedb/src/common/DBInterface.cc#L16).

jsturdy:
@lmoureaux, can you rebase this once #239 is merged?

lmoureaux:
Rebased.

jsturdy:
Oops, I think you were too quick, you rebased on top of #238, not #239

lmoureaux:
Ooops

lmoureaux:
Rebased again.

Issue #234, opened 2019-02-07
[feature] Add HW layout map for configuration via XML


jsturdy:
## Brief summary of issue
Configuration of setups with XML files needs a HW layout map, similar to the extra information that is provided by the DB view

### Types of issue
- [x] Feature request (request for change which adds functionality)

## Context
Needed information is:
* shelf/crate (`AMC`/`VFAT`/`OH`/`GBT`)
* slot (`AMC`/`VFAT`/`OH`/`GBT`)
* link (`VFAT`/`OH`/`GBT`)
* position (`VFAT`/`GBT`)
* unique ID (`chipID`/`s/n`)

PR #238, opened 2019-02-08
Add HW layout object

lmoureaux:
## Description

Add a `SystemTopology` class that describes the HW layout, a `ConfigurationProvider` interface as an abstraction for XML files, and a `ConfigurationLinker` class that creates a tree of configuration objects that corresponds to a given topology.

### Types of changes

## Motivation and Context

See issue #234.

## How Has This Been Tested?

All tests pass.

## Checklist:


gemonlinedb/include/gem/onlinedb/ConfigurationProvider.h
```diff
@@ -0,0 +1,76 @@
...
+         * use with @ref ConfigurationLinker.
+         *
+         * Retrieval functions throw @c std::out_of_range when the requested
+         * configuration is not found.
+         */
+        class ConfigurationProvider
+        {
+        public:
+            /// @brief Constructor.
+            explicit ConfigurationProvider() = default;
```
mexanick: What's the purpose of declaring default constructor with no args as `explicit`?

gemonlinedb/include/gem/onlinedb/ConfigurationProvider.h
```diff
@@ -0,0 +1,76 @@
...
+
+            /// @brief Destructor.
+            virtual ~ConfigurationProvider() = default;
+
+            /**
+             * @brief Retrieves a configuration for the referenced AMC 13
+             *
+             * The corresponding AMC configurations must not be linked.
```
mexanick: "Must not" or "do not have to be"? Same question to line 45, 53 and 61

gemonlinedb/include/gem/onlinedb/DBInterface.h
```diff
@@ -0,0 +1,54 @@
...
+                                                  xdata::Table::Row &row)
+        {
+            auto columns = table.getColumns();
+            detail::RegisterData data;
+            for (const auto &c : columns) {
+                auto value = row.getField(c);
+                if (value != nullptr && value->type() == "string") {
+                    auto str = value->toString();
+                    if (!str.empty()) {
+                        data[c] = std::stoi(value->toString());
```
mexanick: don't call `value->toString()` second time, use temporary var `str` instead

gemonlinedb/include/gem/onlinedb/DBInterface.h
```diff
@@ -0,0 +1,54 @@
...
+                    } else {
+                        // Empty value in database -> use 0
+                        // This *does* actually happen...
```
mexanick: An exception has to be thrown such that caller is notified about DB incompleteness

gemonlinedb/include/gem/onlinedb/DBInterface.h
```diff
@@ -0,0 +1,54 @@
...
+                        data[c] = 0;
+                    }
+                } else {
+                    // "null" in database -> use 0
+                    // This *does* actually happen...
```
mexanick: Same comment as l38

gemonlinedb/include/gem/onlinedb/SystemTopology.h
```diff
@@ -0,0 +1,136 @@
...
+
+namespace gem {
+    namespace onlinedb {
+
+        /**
+         * @brief Describes the topology of a system.
+         *
+         * The description is built as a tree of nodes containing part
+         * references for the various components. The root node is the AMC 13
+         * (only one AMC 13 is supported), under which AMCs are found, then OHs
```
mexanick: We should support multiple AMC13s, as GE1/1 installation at p5 will have two uTCA crates

gemonlinedb/include/gem/onlinedb/SystemTopology.h
```diff
@@ -0,0 +1,136 @@
...
+
+            private:
+                /// @brief Populates a @c VFAT3Node from XML data.
+                void populate(const xercesc::DOMElement *el);
+            };
+
+            /**
+             * @brief Represents an OHv3.
+             */
+            class OHv3Node
```
mexanick: I would try to keep the names as generic as possible. E.g. this code should be reusable by e.g. GE2/1 which is quite similar to GE1/1, but will have different OH and different number of VFATs

gemonlinedb/include/gem/onlinedb/SystemTopology.h
```diff
@@ -0,0 +1,136 @@
...
+            {
+                static const constexpr std::size_t VFAT_COUNT = 24;
```
mexanick: I would pass the number of VFATs as a parameter, so that the code can easily accommodate GE2/1

gemonlinedb/include/gem/onlinedb/SystemTopology.h
```diff
@@ -0,0 +1,136 @@
...
+                /// @brief Populates a @c OHv3Node from XML data.
+                void populate(const xercesc::DOMElement *el);
+            };
+
+            /**
+             * @brief Represents an AMC.
+             */
+            class AMCNode
+            {
+                static const constexpr std::size_t OH_COUNT = 12;
```
mexanick: Same comment as l57

gemonlinedb/include/gem/onlinedb/XMLConfigurationProvider.h
```diff
@@ -0,0 +1,116 @@
...
+
+            /**
+             * @brief Loads OHv3 configurations from an XML file.
+             *
+             * The configurations found in the file are made available using
+             * retrieval functions.
+             *
+             * The file is searched for in the base path (see @ref setSearchPath).
+             */
+            void loadOHv3(const std::string &filename);
```
mexanick: Again, let's be as generic as possible in names

gemonlinedb/src/common/AMC13Configuration.cc
```diff
@@ -134,5 +134,34 @@ namespace gem {
             return dom;
         }
 
+        template<>
+        AMC13Configuration DBInterface::convertRow<AMC13Configuration>(
+            const xdata::Table &table, xdata::Table::Row &row)
+        {
+            AMC13Configuration config;
+
+            auto value = row.getField("FED_ID");
```
mexanick: This method should be refactored: construct a list of fields and iterate over a list calling 
config.SetValue("FieldName", value). A corresponding single method in config has to be implemented instead of the three used here. 
An exception has to be thrown in case value is a `nullptr` or empty.

gemonlinedb/src/common/FileUtils.cc
```diff
@@ -0,0 +1,34 @@
+#include "gem/onlinedb/detail/FileUtils.h"
```
mexanick: move under `src/common/detail/`

gemonlinedb/src/common/SystemTopology.cc
```diff
@@ -0,0 +1,99 @@
...
+            m_root.populate(amc13El);
+        }
+
+        void SystemTopology::AMC13Node::populate(const DOMElement *el)
+        {
+            populatePart(reference, el);
+
+            amc.clear();
+            auto amcEl = el->getFirstElementChild();
+            do {
```
mexanick: here you suppose that the AMC13 node contains at least one AMC child. This has to be checked beforehand and exception thrown if such child node is not found.

gemonlinedb/src/common/XMLConfigurationProvider.cc
```diff
@@ -0,0 +1,203 @@
...
+                XMLSerializationData<ConfigurationType> xmlData;
+                xmlData.readDOM(document);
+
+                auto dataSets = xmlData.getDataSets();
+                for (const auto &set : dataSets) {
+                    auto part = set.getPart();
+                    if (map.count(toString(part)) > 0) {
+                        // Corresponds to more than one configuration for the same
+                        // part, which is not supported.
+                        throw std::runtime_error(
```
mexanick: I would rather define custom exceptions for such things (anyway they ave to be defined for the cases outlined above)

gemonlinedb/include/gem/onlinedb/ConfigurationProvider.h
lmoureaux: I generally mark *all* constructors as explicit. It gives a nice symmetry with the `virtual` destructors.

gemonlinedb/include/gem/onlinedb/ConfigurationProvider.h
lmoureaux: Do not have to be.

gemonlinedb/include/gem/onlinedb/SystemTopology.h
lmoureaux: Ok

gemonlinedb/include/gem/onlinedb/SystemTopology.h
lmoureaux: We can decide to rename it to "OHNode" but I'd do this in a separate PR since there are quite a few things named OHv3* already. Also non-trivial design decisions would need to be made if two separate OH classes are needed.

I fear that supporting both GE2/1 and GE1/1 in the same build would require some heavy refactoring.

gemonlinedb/include/gem/onlinedb/SystemTopology.h
lmoureaux: See discussion [here](https://github.com/cms-gem-daq-project/cmsgemos/pull/231#discussion_r242139340). TL;DR: will be changed to constants from another header, so that a build-time switch between GE1/1 and GE2/1 can be achieved easily.

gemonlinedb/src/common/AMC13Configuration.cc
lmoureaux: A list of fields could only iterate over two of the values since the third is a `string` (which is why this specialization is needed in the first place). I don't think it's worth the additional complexity.

gemonlinedb/src/common/FileUtils.cc
lmoureaux: Such a structure doesn't exist for other `detail` classes. Changing this is probably out of scope here (can be done as another PR to be merged before).

gemonlinedb/src/common/SystemTopology.cc
lmoureaux: Good catch

gemonlinedb/include/gem/onlinedb/DBInterface.h
lmoureaux: Unfortunately it would also prevent me from testing with the existing DB. In addition if new fields are added later they will be initialized to `null` (and would hence break existing data).

gemonlinedb/src/common/AMC13Configuration.cc
mexanick: hmmm... 
```c++
try {
   int val_i = std::stoi(value);
    config.setValue("FieldName", val_i);
} catch (const std::invalid_argument& e) {
    config.setValue("FieldName", value);
}
```

gemonlinedb/include/gem/onlinedb/SystemTopology.h
mexanick: the firmware will hide the hardware changes at max, so I doubt there will be non-trivial design decisions required.

gemonlinedb/include/gem/onlinedb/DBInterface.h
mexanick: Can't you introduce catchers in your test code? The throwing an exception does not always mean an error as it can be intercepted. In the current design a user won't know about flaw in the DB unless explicitly check the values (and 0 could be actually in valid range!)

gemonlinedb/src/common/AMC13Configuration.cc
lmoureaux: * Doesn't handle the case where `value` is a valid number but needs to be set as a string
* Call `setValue(std::string, std::string)` for integer fields when the integer isn't valid and it should fail
* The functions `config.setValue(std::string, X)` would basically contain a switch statement delegating to the individual functions already called in the current implementation, which would add one layer for very little gain

gemonlinedb/include/gem/onlinedb/SystemTopology.h
lmoureaux: I feared for a moment that I'd need to refactor the 2,000 lines already committed to this branch... I still think that the naming convention shouldn't be changed in this PR.

gemonlinedb/include/gem/onlinedb/DBInterface.h
lmoureaux: Since only the AMC configuration is affected I disabled the corresponding test.

gemonlinedb/src/common/AMC13Configuration.cc
lmoureaux: Somewhat cleaner with a macro: https://github.com/cms-gem-daq-project/cmsgemos/pull/237/files#diff-5e87771af216d4fce8dfa4eb565b7070R144

mexanick:
Why should we have cases where a valid integer is not stored as `int` and must be stored as `string`?

On 8 Feb 2019, at 23:18, lmoureaux <notifications@github.com<mailto:notifications@github.com>> wrote:


@lmoureaux commented on this pull request.

________________________________

In gemonlinedb/src/common/AMC13Configuration.cc<https://github.com/cms-gem-daq-project/cmsgemos/pull/238#discussion_r255253469>:

> @@ -134,5 +134,34 @@ namespace gem {
             return dom;
         }

+        template<>
+        AMC13Configuration DBInterface::convertRow<AMC13Configuration>(
+            const xdata::Table &table, xdata::Table::Row &row)
+        {
+            AMC13Configuration config;
+
+            auto value = row.getField("FED_ID");


  *   Doesn't handle the case where value is a valid number but needs to be set as a string
  *   Call setValue(std::string, std::string) for integer fields when the integer isn't valid and it should fail
  *   The functions config.setValue(std::string, X) would basically contain a switch statement delegating to the individual functions already called in the current implementation, which would add one layer for very little gain

lmoureaux:
Added missing XML files and tests. Removed all DB commits since they proved to be confusing (handled in #237).

gemonlinedb/src/common/XMLConfigurationProvider.cc
jsturdy: Yes, all `xdaq` related code should throw (a derivative of) `xcept:exception`, you can use the on-the-fly exception creation macros in the, e.g., `include/gem/onlindb/exception/Exception.h` header, or define specific exceptions there.

jsturdy:
Merging blocked by #237 for sequential (better in future to branch from same parent, and rebase once other features have merged, unless strictly dependent on everything in a separate feature PR)

gemonlinedb/src/common/XMLConfigurationProvider.cc
lmoureaux: Already fixed in 8562f2918b694c9ce6fb1c809b6c043e43af0ee7

gemonlinedb/src/common/AMC13Configuration.cc
lmoureaux: Fixed in 05feeefcc80c0dd45e0038a82c8c97e45fabc0e9 (#237)

lmoureaux:
> Merging blocked by #237 for sequential

Nope, commits in #237 were removed from this branch.

gemonlinedb/.gitignore
```diff
@@ -3,6 +3,9 @@
 
 # Don't ignore common xsd definitions
 !xml/schema/common.xsd
+!xml/schema/system-topology.xsd
+!xml/examples/*.xml
+!xml/*.xml
```
jsturdy: Any reason to not ignore all `.xml`/`.xsd` files in these locations?
```
!xml/**/*.xsd
!xml/**/*.xml
```

gemonlinedb/xml/schema/system-topology.xsd
```diff
@@ -0,0 +1,79 @@
...
+    <xs:element name="vfat3">
+        <xs:complexType>
+            <xs:attribute ref="gem:barcode" use="required"/>
+        </xs:complexType>
+    </xs:element>
+
+    <!-- OHv3 -->
+    <xs:element name="ohv3">
+        <xs:complexType>
+            <xs:choice minOccurs="24" maxOccurs="24">
```
jsturdy: Could have fewer, depending on the use case, or does the `empty-slot` cover cases when there are fewer actual `VFATs` connected?

lmoureaux: That's what `<empty-slot>` does, yes. Entries are expected to be sequential. I should probably add documentation comments to the examples.

gemonlinedb/.gitignore
lmoureaux: `xsd` because some of them are generated. Could ignore all `xml` indeed.

Issue #241, opened 2019-02-13
[feature][gemonlinedb] Build and run tests on CI


lmoureaux:
## Brief summary of issue

`gemonlinedb` comes with a set of tests. It's a pity that CI doesn't use them.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

Tests are run on CI, and merging of PR that break tests is blocked.

## Current Behavior

Tests aren't run on CI.

## Possible Solution (for bugs)

Add CI stages that build and run tests. Note that some of them may open a connection to the DB, I'm not sure this can be done on Travis.

## Context

Merging of PR #239 that resulted in issue #240 could have been prevented by CI.

jsturdy:
> Note that some of them may open a connection to the DB, I'm not sure this can be done on Travis.

Unlikely, but will probably be possible in the `gitlab-ci` builds.
Can you separate the tests that require a DB connection, from those that test other features?

May require a new top-level `tests` target and associated rules

jsturdy:
Minimally changed (just to allow compilation) done in #242

cgalloni:
@lmoureaux , can this be closed/or ported to gitlab?

lpetre-ulb:
The tests run and must pass in the new repository CI.

PR #243, opened 2019-02-13
[feature] Update build system to be able to run tests on CI

lmoureaux:
## Description

This PR adds support for running tests from the top-level `Makefile`. There are two targets:

* `run-tests`: Runs all tests
* `run-tests-ci`: Runs all tests except the ones requiring special setup such as a working DB 

### Types of changes

## Motivation and Context

See #241 

## How Has This Been Tested?

* Tests can be run from the top of the source tree
* Failing tests are reported to the caller using a non-zero return status

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,13 +108,13 @@ $(SUBPACKAGES.INSTALL):
 $(SUBPACKAGES.CLEAN):
 	$(MAKE) -C $(patsubst %.clean,%, $@) clean
 
-$(SUBPACKAGES.TESTS):
+$(SUBPACKAGES.TESTS): all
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -84,6 +86,10 @@ clean: $(SUBPACKAGES.CLEAN)
 
 tests: $(SUBPACKAGES) $(SUBPACKAGES.TESTS)
 
+run-tests: $(SUBPACKAGES.RUNTESTS)
```
jsturdy: These both depend on having the `tests` target run, right?

lmoureaux: Indeed :P

lmoureaux:
Fixed a bug and dependencies.

PR #263, opened 2019-03-20
Add link status check to test connectivity

BenjaminRS:
## Description

Adding the ability to check for the trigger link status. Therefore needed to add in extra functionality to the xhal.

### Types of changes

## Motivation and Context

This PR addresses #224 and #225.

## How Has This Been Tested?

Tested various different setups of the testConnectivity.py options where, for example, the link would fail.

## Checklist:

### DIFF
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -7,6 +7,7 @@ from reg_utils.reg_interface.common.reg_xml_parser import getNode, parseInt, par
 from xhal.reg_interface_gem.core.reg_extra_ops import rBlock
 
 import logging
+import os
 
 gMAX_RETRIES = 5
 gRetries = 5
@@ -127,6 +128,10 @@ class HwAMC(object):
         self.getmonOHLink = self.lib.getmonOHLink
         self.getmonOHLink.argTypes = [ OHLinkMonitorArrayType, VFATLinkMonitorArrayType, c_uint, c_uint, c_bool ]
         self.getmonOHLink.restype = c_uint
+
+        self.getmonTRIGGEROHmain = self.lib.getmonTRIGGEROHmain
+        self.getmonTRIGGEROHmain.argTypes = [ POINTER(c_uint32), c_uint, c_uint ]
+        self.getmonTRIGGEROHmain.restype = c_uint
 
         self.getmonVFATLink = self.lib.getmonVFATLink
         self.getmonVFATLink.argTypes = [ VFATLinkMonitorArrayType, c_uint, c_uint, c_bool ]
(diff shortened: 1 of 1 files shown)


gempython/tools/amc_user_functions_xhal.py
```diff
@@ -401,6 +406,71 @@ def getTTCStatus(self, ohN, display=False):
                 print("\tENABLE: \t\t\t%i"%(running))
         return running
 
+    def getTriggerLinkStatus(self,printSummary=False, checkCSCTrigLink=False, ohMask=0xfff):
+        """
+        Gets the trigger link status
+        """
```
bdorney, line 419: The documentation here should describe how this will behave and how `ohMask` will be treated in each of the cases of `checkCSCTrigLink`.  See for example lines 474 to 484.

bdorney (changes requested):
Method documentation needs expansion; otherwise good.

gempython/tools/amc_user_functions_xhal.py
BenjaminRS: ok I have updated the text here -- please have a look at this one again

jsturdy:
As you will need to do a `rebase`, can you cleanup your commit history?
* `squash` cleanup commits into single commit, or `fixup`
* `squash` (and potentially reorder) related feature commits with their "fixup" commits
e.g.,
```
pick * 90e8ce2      Adding test link -- currently in debugging state
pick * ea5754b      Moving code to xhal
fixup * 3962570      Cleaning up code
fixup * 37dbdcb      Cleaning up code and making the output pretty
pick * f19ae05      Adding switch to write out phase settings
pick * b48befc      adding switch for CSC
squash * 86a178b    Updating to move checkCSCTrigLink to the command line
```

jsturdy (commented):
As you will need to do a rebase, can you cleanup your commit history?

* `squash` cleanup commits into single commit, or `fixup`
* `squash` commits that fix a bug introduced during feature development
* `squash` (and potentially reorder) related feature commits with their "fixup" commits

BenjaminRS:
updated this and [vfatqc-python-scripts#232](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/232) to squash the commits

gempython/tools/amc_user_functions_xhal.py
bdorney: Lines 411 and 412 are fine.  But subsequent lines I think are still not quiet clear. I would say:

> ohMask is a 12 bit number, see checkCSCTrigLink for how each bits are interpreted.
> If checkCSCTrigLink is False the ohMask will define which OH's to query, a 1 in the N^th bit means to query the N^th optohybrid.
> If checkCSCTrigLink is True, the ohMask is expected to have all odd bits set to 0.  Only even bits may be nonzero.  If the N^th even bit is nonzero the trigger link status for that OHN and OHN+1 will be queried.  It is expected that the GEM trigger link from the physical optohybrid is going to the OHN fiber slot on this AMC and the CSC Trigger Link from the physical optohybrid is going to the OHN+1 fiber slot on this AMC. 

I think this would be more representative and clear.

bdorney (changes requested):
Slight tweak to the documentation; also probably another round of `squash` may be needed.

jsturdy:
> also probably another round of `squash` may be needed.

Just to be more clear, as I was probably unclear with my request, the goal of the commit cleanup is not to simply have one commit per PR, but rather to squash intermediate commits that a developer makes locally that are (mostly) irrelevant to the end result.
If you are working from a "commit early, commit often" mindset (and I am not advocating against this at all), chances are that one makes many changes while working out some particular aspect of their work, and then finally has it working correctly.  It is this finally working correctly thing that is really being part of the PR, so all those intermediate things (like, "oops, missed a newline", or "forgot to add new parameter") should be squashed out.

I'll try to work up a concrete example for the next developers meeting

bdorney:
> > also probably another round of `squash` may be needed.
> 
> Just to be more clear, as I was probably unclear with my request, the goal of the commit cleanup is not to simply have one commit per PR, but rather to squash intermediate commits that a developer makes locally that are (mostly) irrelevant to the end result.
> If you are working from a "commit early, commit often" mindset (and I am not advocating against this at all), chances are that one makes many changes while working out some particular aspect of their work, and then finally has it working correctly. It is this finally working correctly thing that is really being part of the PR, so all those intermediate things (like, "oops, missed a newline", or "forgot to add new parameter") should be squashed out.
> 
> I'll try to work up a concrete example for the next developers meeting

Damn, I should have saved the git cleanup I just did...is there a way to expand squashed commits? I can show the before and after and the reasoning behind my squashing.

BenjaminRS:
Made a new commit to put in @bdorney suggestion for the description

bdorney (approved):

jsturdy (approved):
Cursory

PR #266, opened 2019-03-23
[feature] OnlineDB ConfigurationManager

lmoureaux:
## Description

Add a new class ConfigurationManager as the main entry point for gemonlinedb. Use the opportunity to better define the exception and thread safety policies within gemonlinedb.

### Types of changes

## Motivation and Context

Requested by Jared to implement hardware objects.

## How Has This Been Tested?

Unit tests.

## Checklist:

### DIFF
--- a/gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
+++ b/gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
@@ -9,6 +9,8 @@
 #include <xercesc/sax/SAXException.hpp>
 #include <xercesc/util/XMLString.hpp>
 
+#include "gem/onlinedb/exception/Exception.h"
+
 /**
  * @file
  * @brief Defines utilities to make Xerces slightly less verbose.
@@ -143,6 +145,50 @@ namespace gem {
                 void resetErrors() override {};
             };
 
+            /**
+             * @brief Turns Xerces exceptions thrown by a function or functor
+             *        object into @ref ParseError.
+             */
+            template<class F, class... Args>
+            auto xercesExceptionsToXcept(F f, Args... args) ->
+                decltype(f(std::forward<Args>(args)...))
+            {
+                XERCES_CPP_NAMESPACE_USE
+                using exception::ParseError;
+
+                XercesGuard guard;
+                try {
+                    return f(std::forward<Args>(args)...);
+                } catch(DOMException &e) {
+                    XCEPT_RAISE(ParseError, transcode(e.getMessage()));
+                } catch (XMLException &e) {
+                    XCEPT_RAISE(ParseError, transcode(e.getMessage()));
+                } catch (SAXParseException &e) {
+                    XCEPT_RAISE(ParseError,
+                        transcode(e.getPublicId()) + ": " +
+                        std::to_string(e.getLineNumber()) + ":" +
+                        std::to_string(e.getColumnNumber()) + ": " +
+                        transcode(e.getMessage()));
+                } catch (SAXException &e) {
+                    XCEPT_RAISE(ParseError, transcode(e.getMessage()));
+                }
+            }
+
+            /**
+             * @brief Turns Xerces exceptions thrown by a function or functor
+             *        object into @ref std::runtime_error.
+             */
+            template<class F, class... Args>
+            auto xercesExceptionsToStd(F f, Args... args) ->
+                decltype(f(std::forward<Args>(args)...))
+            {
+                try {
+                    return xercesExceptionsToXcept(f, std::forward<Args>(args)...);
+                } catch(xcept::Exception &e) {
+                    // Use the opportunity to convert all Xcept exceptions
+                    throw std::runtime_error(e.message());
+                }
+            }
 
             /**
              * @brief Appends a child element to @c parent with tag name
--- a/gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
+++ b/gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
@@ -2,6 +2,7 @@
 #define GEM_ONLINEDB_DETAIL_XMLUTILS_H
 
 #include <memory>
+#include <mutex>
 #include <string>
 
 #include <xercesc/dom/DOM.hpp>
@@ -115,6 +116,8 @@ namespace gem {
              */
             class XercesGuard
             {
+                static std::recursive_mutex s_mutex;
+
             public:
                 /**
                  * @brief Initializes Xerces.
--- a/gemonlinedb/src/common/ConfigurationManager.cc
+++ b/gemonlinedb/src/common/ConfigurationManager.cc
@@ -1,0 +1,128 @@
+#include "gem/onlinedb/ConfigurationManager.h"
+
+#include <cstdlib>
+
+#include <xercesc/dom/DOM.hpp>
+#include <xercesc/parsers/XercesDOMParser.hpp>
+
+#include "gem/onlinedb/ConfigurationLinker.h"
+#include "gem/onlinedb/SystemTopology.h"
+#include "gem/onlinedb/XMLConfigurationProvider.h"
+#include "gem/onlinedb/detail/FileUtils.h"
+
+namespace gem {
+    namespace onlinedb {
+
+        namespace /* anonymous */ {
+
+            /**
+             * \brief Loads a system-topology.xml file
+             */
+            DOMDocumentPtr loadXMLTopologyFile(const std::string &path)
+            {
+                using namespace detail::literals;
+                XERCES_CPP_NAMESPACE_USE
+
+                detail::XercesGuard guard; // Make sure that Xerces is loaded
+
+                // Find the schema (throws if not found)
+                auto schemaPath = detail::getFileInPath(
+                    "system-topology.xsd",
+                    "xml/schema:/opt/cmsgemos/share/gemonlinedb/xml/schema");
+
+                // Find the file (throws if not found)
+                auto filePath = detail::getFileInPath("system-topology.xml", path);
+
+                // Get an implementation
+                auto impl = DOMImplementationRegistry::getDOMImplementation("LS"_xml);
+                if (impl == nullptr) {
+                    throw std::logic_error("Cannot find a DOM implementation");
+                }
+
+                auto errorHandler =
+                    std::make_shared<detail::XercesAlwaysThrowErrorHandler>();
+
+                XercesDOMParser parser;
+                parser.setErrorHandler(errorHandler.get());
+                parser.setExternalSchemaLocation(
+                    ("https://github.com/cms-gem-daq-project/cmsgemos/tree/master/gemonlinedb/xml/schema/system-topology.xsd "
+                    + schemaPath).c_str());
+                parser.setValidationScheme(XercesDOMParser::Val_Always);
+                parser.setDoNamespaces(true);
+                parser.setDoSchema(true);
+
+                // Parse and validate
+                parser.parse(detail::XercesString(filePath));
+
+                // Load contents
+                DOMDocumentPtr document;
+                document.reset(parser.adoptDocument());
+                return document;
+            }
+
+        } // anonymous namespace
+
+        std::unique_ptr<ConfigurationManager>
+        ConfigurationManager::s_instance = nullptr;
+
+        ConfigurationManager::ConfigurationManager(Source objectSource,
+                                                   Source topologySource)
+        {
+            // Deduce the search path for XML files
+            std::string path = "/etc/opt/cmsgemos"; // FHS 3.0 standard location
+            if (auto envPath = std::getenv("CMSGEMOS_CONFIG_PATH")) {
+                path = envPath;
+            }
+
+            ConfigurationLinker linker;
+
+            // Configuration objects
+            switch (objectSource) {
+            case Source::DB:
+                // TODO Not supported yet
+                break;
+
+            case Source::XML:
+                auto provider = std::make_shared<XMLConfigurationProvider>();
+                provider->setSearchPath(path);
+                provider->load(loadXMLTopologyFile(path));
+                linker.setProvider(provider);
+                break;
+            }
+
+            // Topology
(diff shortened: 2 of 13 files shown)


gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
```diff
@@ -143,6 +148,50 @@ namespace gem {
                 void resetErrors() override {};
             };
 
+            /**
+             * @brief Turns Xerces exceptions thrown by a function or functor
+             *        object into @ref ParseError.
+             */
+            template<class F, class... Args>
+            auto xercesExceptionsToXcept(F f, Args... args) ->
```
mexanick, line 156: Neatpicking here: avoid if possible returning `auto` types

gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
```diff
@@ -143,6 +148,50 @@ namespace gem {
...
+                    XCEPT_RAISE(ParseError, transcode(e.getMessage()));
+                }
+            }
+
+            /**
+             * @brief Turns Xerces exceptions thrown by a function or functor
+             *        object into @ref std::runtime_error.
+             */
+            template<class F, class... Args>
+            auto xercesExceptionsToStd(F f, Args... args) ->
```
mexanick, line 185: same as above

gemonlinedb/src/common/ConfigurationManager.cc
```diff
@@ -0,0 +1,167 @@
...
+
+                detail::XercesGuard guard; // Make sure that Xerces is loaded
+
+                // Find the schema (throws if not found)
+                auto schemaPath = detail::getFileInPath(
+                    "system-topology.xsd",
+                    "xml/schema:/opt/cmsgemos/share/gemonlinedb/xml/schema");
+
+                // Find the file (throws if not found)
+                auto filePath = detail::getFileInPath("system-topology.xml", path);
```
mexanick, line 34: is this `std::string` or `char*`? If so, use explicit type declaration instead of `auto`

gemonlinedb/src/common/ConfigurationManager.cc
```diff
@@ -0,0 +1,167 @@
...
+                    throw std::logic_error("Cannot find a DOM implementation");
+                }
+
+                auto errorHandler =
+                    std::make_shared<detail::XercesAlwaysThrowErrorHandler>();
+
+                XercesDOMParser parser;
+                parser.setErrorHandler(errorHandler.get());
+                parser.setExternalSchemaLocation(
+                    ("https://github.com/cms-gem-daq-project/cmsgemos/tree/master/gemonlinedb/xml/schema/system-topology.xsd "
```
mexanick, line 50: Should be a `const std::string` defined somewhere in a central location rather than hardcoded here (e.g. we can move the repo to gitlab etc.). Also a note: such code requires external `https` access which is at least cumbersome at P5 and might be not recommended way to go...

gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
lmoureaux: Let me know if I'm wrong, but I don't think it is actually possible here.

gemonlinedb/src/common/ConfigurationManager.cc
lmoureaux: If it were `char*` I would have changed `getFileInPath()` to return `std::string` ;)

gemonlinedb/src/common/ConfigurationManager.cc
lmoureaux: This syntax defines the location of the XML schema identified by `https://github.com/cms-gem-daq-project/...`. The second argument (after the whitespace) is its path in the local filesystem. No remote access is required.

The schema URL has to be written in XML files (when they are namespace-aware like this one), the `xsd` file and the loading code. It is hence part of the data format.

In the event of a migration to Gitlab, I'm hoping that a read-only archive would be kept for historical purpose.

gemonlinedb/src/common/ConfigurationManager.cc
bdorney: I think @mexanick's point was to explicitly illustrate the type and not use `auto`.

lmoureaux: What would be a reason not to use `auto` as recommended by many C++ 11 tutorials?

jsturdy:
My personal view regarding `auto`/`decltype` can be summarized as follows:
I don't care at all when they show up in the definition/implementation.
I've been reluctant to (over)use them in headers for the primary reason that explicit typing is useful as a documentation tool.

I don't want to be too much of a curmudgeonly old man, so I'll just say this:
* If it doesn't obfuscate usage, go for it
* If it aids the readability, go for it
* If it doesn't clearly satisfy the aforementioned points, ask if/why you think it's necessary

gemonlinedb/src/common/ConfigurationManager.cc
```diff
@@ -0,0 +1,167 @@
...
+
+        boost::shared_mutex ConfigurationManager::s_mutex;
+
+        ConfigurationManager::ConfigurationManager(Source objectSource,
+                                                   Source topologySource)
+        {
+            try {
+                detail::xercesExceptionsToXcept([&]{
+                    // Deduce the search path for XML files
+                    std::string path = "/etc/opt/cmsgemos"; // FHS 3.0 standard location
```
jsturdy: `/etc/opt/cmsgemos.d` is created by the RPM build process, with an optional `/etc/opt/cmsgemos.conf` as some sort of package configuration probably
I might prefer `/opt/cmsgemos/etc/...` as a location (everything can then be referenced from `CMSGEMOS_ROOT`), but in the end, if it's packaged correctly, it won't matter

jsturdy (changes requested):

gemonlinedb/src/common/ConfigurationManager.cc
lmoureaux: Referring to [section 3.7.4](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html#etcoptConfigurationFilesForOpt) of the Filesystem Hierarchy Standard, the path should be `/etc/opt/...`. The reasoning behind the standard seems to be that the complete configuration from a given host resides in one directory that can be backed up independently, or copied over to a clean install.

Files in the config path are host-specific in nature (and optional: XML will not be used at P5) so the packages do not include them. The intent is that sysadmins will maintain a set of files matching their setup in `/etc/opt/cmsgemos`. When users need to override (some of) them, they can set the environment variable with the same syntax as a traditional `PATH`.

`.d` directories usually have a special meaning: configuration fragments found in these directories will be merged together to produce the final config. `gemonlinedb` doesn't support this method of merging. Since `/etc/opt/cmsgemos.d` appears not to be used at all, I would  replace it by `/etc/opt/cmsgemos`. Most directories in `/etc` do not use the `.d` suffix anyway.

jsturdy: The close linking between the build process for `cmsgemos` and `xdaq` would encourage adoption of a similar structure
```
[]% ll /opt/xdaq/
total 40
drwxr-xr-x.  2 root root  4096 Feb  1  2018 bin
drwxr-xr-x.  2 root root  4096 Feb  1  2018 config
drwxr-xr-x.  3 root root  4096 Feb  1  2018 etc
drwxr-xr-x. 50 root root  4096 Feb 14 15:13 htdocs
drwxr-xr-x. 59 root root  4096 Feb  1  2018 include
drwxr-xr-x.  4 root root 12288 Feb 22 12:15 lib
drwxr-xr-x.  2 root root  4096 Dec 16  2016 sbin
drwxr-xr-x.  4 root root  4096 Oct 25  2017 share
[]% ll /opt/xdaq/etc
total 48
-rw-r-xr-x. 1 root root  1826 Nov 14  2017 base.profile
drwxr-xr-x. 5 root root  4096 Dec 19  2016 CAENUpgrader
-rw-r-xr-x. 1 root root  3909 Nov 14  2017 default.profile
-rw-r-xr-x. 1 root root  1899 Nov 14  2017 default.profile.macosx
-rw-r-xr-x. 1 root root   462 Dec 16  2016 htaccess
-rw-r-xr-x. 1 root root    53 Dec 16  2016 htgroup
-rw-r-xr-x. 1 root root    19 Dec 16  2016 htpasswd
-rw-r-xr-x. 1 root root   255 Dec 16  2016 log4cplus.properties
-rwxr-xr-x. 1 root root 10765 Dec 16  2016 tinyproxy.conf
-rw-r-xr-x. 1 root root   435 Dec 16  2016 udpServer.sh
[]% ll /opt/cactus
total 16
drwxr-xr-x. 4 root root 4096 Aug  1  2018 bin
drwxr-xr-x. 4 root root 4096 Aug  1  2018 etc
drwxr-xr-x. 4 root root 4096 Aug  1  2018 include
drwxr-xr-x. 3 root root 4096 Aug  3  2018 lib
[]% ll /opt/cactus/etc
total 8
drwxr-xr-x. 2 root root 4096 Jan 31  2018 amc13
drwxr-xr-x. 4 root root 4096 Aug  1  2018 uhal
[]% ll /etc/opt
total 8
drwxr-x---. 2 gempro gempro 4096 Feb 14 15:55 cmsgemos.d
drwxr-xr-x. 9 root   root   4096 Mar 24  2018 rh
```

Though, on the other hand, the `xdaq` zone configurations are installed in
```
/etc/xdaq.d
```

So, given that
```
/opt/cmsgemos/etc
```
and 
```
/etc/opt/cmsgemos.d
```
are already created (and at least one of them used), unless there is a reasonable concern that you're attempting to resolve, one of the two should be used.

jsturdy (commented):
Last thing, there is a mix of `doxygen` syntax styles here, that we should normalize
I think I've used `@tag` everywhere in this repo (others of our repos have adopted `\tag`), and in this PR you have a mix of both.

lmoureaux:
Fixed.

gemonlinedb/src/common/ConfigurationManager.cc
mexanick: Again, I would define it as a string constant in a well-known place, rather than burried somewhere in the code, so in case we need to change it, it would be more straightforward

gemonlinedb/src/common/ConfigurationManager.cc
mexanick: The reason is that in this case it is not obvious what's the returned type (e.g. contrary to nearby constructions like `make_shared...`

jsturdy (commented):
I lied, one more :-)
The member values prefixed by `///<` don't require any further tags, I think `doxygen` gets them right automatically
Then can you squash all of these commits (related to `doxygen` syntax)

gemonlinedb/src/common/ConfigurationManager.cc
lmoureaux: Then give me a `configure` step to sync it with the `xsd` and example `xml`… I can do that with any modern build system, but not the xDAQ Makefiles.

Note that changing this value would result in old files not working anymore (that's what I meant with "part of the data format").

gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
mexanick: Ah, I missed that line 164 in cursory look. Then, fine. Sometimes in the future should move to c++14 (maybe in 10 years :D)

gemonlinedb/src/common/ConfigurationManager.cc
jsturdy: Jumping in late here, I wouldn't set the schema URL to the VC file, rather to a site, e.g., where we'll host, e.g., the documentation and `yum` repos (yet to be fully configured, almost certainly somewhere in https://www.cern.ch/cmsgemdaq/sw/)
Then it can be "versioned" in the event the schema itself needs to be updated.
Additionally, as it's currently just a namespace identifier, I think we can leave it as is for now (but @lmoureaux can you add a `FIXME host for schema` comment next to it so my routine searches for `FIXME` will ensure this doesn't get lost?)

jsturdy:
I meant these should be squashed into a single commit (though I should probably wait for the result of your "to be squashed" comment :-) )
```
* 41e43e7        (louis/feature/onlinedb-manager) More dox fixes to be squashed
* 1742871        ConfigurationManager.cc: \brief -> @brief
* 8cf4260        ConfigurationManager.h: \c -> @c
* bacb362        ConfigurationManager.h: \brief -> @brief
```

gemonlinedb/src/common/ConfigurationManager.cc
lmoureaux: You wouldn't know the actual type being returned, only that it is compatible with `std::string`. Now consider the following snippets:
#### 1. No auto
```cpp
char *foo();
std::string bar = foo();
std::cout << bar.size() << " " << (bar + "coming from foo") << std::endl;
```
#### 2. auto
```cpp
char *foo();
auto bar = foo();
std::cout << bar.size() << " " << (bar + "coming from foo") << std::endl;
```
Both have a potential memory leak. Number 1 is perfectly valid code that compiles without warning. Number 2 fails on line 3 with an error (`bar.size()`) and, with recent compilers, a warning (`bar + "coming from foo"` doesn't do what you think).

gemonlinedb/include/gem/onlinedb/detail/XMLUtils.h
lmoureaux: Actually xDAQ 15 should bring C++14 support.

lmoureaux:
Doxygen commits squashed.

jsturdy (approved):

gemonlinedb/src/common/ConfigurationManager.cc
mexanick: Ok, I rest my case that it would be better (more descriptive) to use an `std::string` return type here, but that's minor.

mexanick (approved):

PR #275, opened 2019-04-18
 Gemcalibration package second PR

cgalloni:
### Types of changes

## Motivation and Context

Implementation of the XDAQ web interface

## How Has This Been Tested?

Code tested on gem904daq02

## Checklist:


setup/etc/profile.d/gemdaqenv.sh
```diff
@@ -9,7 +9,7 @@ if [ "$XDAQ_ROOT" != "0" ]; then
     export XDAQ_ROOT=/opt/xdaq
     export XDAQ_DOCUMENT_ROOT=$XDAQ_ROOT/htdocs
     # only do this for regular users
-    # export XDAQ_DOCUMENT_ROOT=/data/xdaq/${USER}
```
mexanick, line 12: This change must be reverted, as it is only required for development and not suitable for releasing

mexanick (changes requested):

setup/etc/profile.d/gemdaqenv.sh
cgalloni: done

Makefile
```diff
@@ -11,6 +11,7 @@ SUBPACKAGES := \
         gempython \
         gemdaqmonitor \
         gemonlinedb \
+        gemcalibration \
         # gemHwMonitor \
```
jsturdy, line 16: Missing `gemcalibration` rule in top-level `Makefile`

config/mfPythonDefsGEM.mk
```diff
@@ -9,6 +9,10 @@ PYTHON_LIB     ?= python${PYTHON_VERSION}
 # Python Config
 PYTHONCFLAGS = $(shell pkg-config python --cflags)
 PYTHONLIBS   = $(shell pkg-config python --libs)
+PYTHONGLIBS  = $(shell pkg-config python --glibs)
+
+IncludeDirs+=${PYTHON_INCLUDE_PREFIX}
+
```
jsturdy, line 12: Where did these get added from?

gemcalibration/include/gem/calib/CalibrationWeb.h
```diff
@@ -0,0 +1,91 @@
...
+#include "gem/calib/Calibration.h"
+
+namespace gem {
+  namespace calib {
+
+    class Calibration;
+
+    class CalibrationWeb: public gem::base::GEMWebApplication
+      {
+        // friend class GEMMonitor;
```
jsturdy: Remove unnecessary commented lines

gemcalibration/src/common/Calibration.cc
```diff
@@ -0,0 +1,226 @@
...
+            } // end loop over NAMC 
+        } // end if checked for shelf
+    } //end loop over shelves
+    //TODO: should we check if at least one link is selected??
+    
+
+    //DACSCAN type
+    if (m_calType==DACSCANV3){
+      std::map<dacScanType_t, std::map<std::string, uint32_t> >::iterator it;
+      for(it=m_dacScanTypeParams.begin();it!=m_dacScanTypeParams.end();it++){
```
jsturdy: * whitespace should follow the form `for (blah) {` (@lmoureaux has added many of my style requirements to the `CONTRIBUTING.md` guidelines)
* Use scoped range based loops with `auto &` or `auto const&` as necessary rather than declaring an iterator outside the loop (unless you have a valid reason not to)

gemcalibration/src/common/Calibration.cc
```diff
@@ -0,0 +1,226 @@
...
+	t_stream.clear();
+	t_stream.str(std::string());
+	t_stream << m_dacScanTypeParams_label.find(it->first)->second ;
+	bool checked = false;
+	checked = cgi.queryCheckbox(t_stream.str());
+	if (checked) {
+	  std::map<std::string, uint32_t> t_dacScan_parameters = it->second;
+	  for (auto dacScan_parameter: t_dacScan_parameters) {
```
jsturdy: again, scope variables unless you have a valid reason not to
```c++
for (auto & par : it->second)
```

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,394 @@
...
+  *out << cgicc::script().set("type", "text/javascript")
+    .set("src", "/gemdaq/gemcalibration/html/scripts/cal_routines.js") 
+       << cgicc::script() << std::endl;
+  
+  *out << "<link  rel=\"stylesheet\" href=\"/gemdaq/gembase/html/css/bootstrap.css\" type=\"text/css\">" << std::endl;
+  *out << "<link  rel=\"stylesheet\" href=\"/gemdaq/gembase/html/css/bootstrap.min.css\" type=\"text/css\">" << std::endl;
+  *out << "<script src=\"/gemdaq/gembase/html/scripts/bootstrap.js\"></script>" << std::endl;
+  *out << "<script src=\"https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js\"></script>" << std::endl;
+
+  *out << "<div class=\"xdaq-tab-wrapper\">" << std::endl;
```
jsturdy: Formatting in this file *must* be for `c++`, not `html`.
If you also want formatted `html` in the output, ensure what you write to the stream is formatted as desired.

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,394 @@
...
+    CMSGEMOS_DEBUG("CalibrationWeb::calibrationPage");
+    *out << "<div class=\"xdaq-tab-wrapper\">" << std::endl;
+        *out << "<div align=\"center\">" << std::endl;
+
+            *out<< "<form id=\"cal_select\">"<<std::endl; *out << "<div class=\"form-group row\">" << std::endl;
+                    *out << "<h2><span class=\"label label-info col-md-3\" id=\"cal_scurve\">Select calibration type:" << "</span></h2>" << std::endl; 
+                   *out << "<div class=\"col-md-3\">" << std::endl;
+
+                    *out << "<select class=\"form-control form-control-lg\" id=\"cal_type_select\" name=\"cal_type_select\" onChange=\"selectCalType()\">" << std::endl;
+                        *out << "<option disabled selected value> -- select an option -- </option>" << std::endl;
```
jsturdy: Same comment throughout this block re formatting

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,394 @@
...
+                                            << "</div>"
+                                        << "</span>" << std::endl;
+                                    } // end loop over NAMC 
+                                *out << "</div>"
+                            << "</div>" // end <div class="dropdown">
+                        << "</div>" // end <div class="col">
+                    << "</div>" << std::endl; // end <div class="row">
+                } // end loop over NSHELF
+            *out << "</div>" << std::endl; // end container
+        *out << "</form>"<< std::endl;
```
jsturdy: This entire file should be reformatted, and subsequently, all trailing whitespace removed, before I can perform further review.

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+#include "gem/base/GEMApplication.h"
+#include "gem/utils/Lock.h"
+#include "gem/utils/LockGuard.h"
+
+//#include "gem/daqmon/exception/Exception.h"
+#include "gem/utils/exception/Exception.h"
+//#include "gem/daqmon/DaqMonitor.h"
+
+#define NAMC 12
+#define NSHELF 2
```
jsturdy: For sure, this should not be hardcoded, take it as a parameter from the xml config

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+
+#include "gem/base/GEMApplication.h"
+#include "gem/utils/Lock.h"
+#include "gem/utils/LockGuard.h"
+
+//#include "gem/daqmon/exception/Exception.h"
+#include "gem/utils/exception/Exception.h"
+//#include "gem/daqmon/DaqMonitor.h"
+
+#define NAMC 12
```
jsturdy: Use `gem::base::GEMApplication::MAX_AMCS_PER_CRATE`

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+
+
+namespace gem {
+  namespace calib {
+
+    enum calType {NDEF, GBTPHASE, LATENCY, SCURVE, SBITARMDACSCAN, ARMDACSCAN, TRIMDAC, DACSCANV3, CALIBRATEARMDAC}; 
+    typedef enum calType calType_t;
+
+
+    enum dacScanType {CFG_CAL_DAC, CFG_BIAS_PRE_I_BIT, CFG_BIAS_PRE_I_BLCC, CFG_BIAS_PRE_I_BSF, CFG_BIAS_SH_I_BFCAS, CFG_BIAS_SH_I_BDIFF, CFG_BIAS_SD_I_BDIFF, CFG_BIAS_SD_I_BFCAS, CFG_BIAS_SD_I_BSF, CFG_BIAS_CFD_DAC_1, CFG_BIAS_CFD_DAC_2, CFG_HYST, CFG_THR_ARM_DAC, CFG_THR_ZCC_DAC, CFG_BIAS_PRE_VREF, CFG_VREF_ADC};
```
jsturdy: Likewise, and document

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+        void pauseAction(xgi::Input *in, xgi::Output *out)
+          throw (xgi::exception::Exception);
+	
+        void applyAction(xgi::Input *in, xgi::Output *out)
+          throw (xgi::exception::Exception);
+        
+        void setCalType(xgi::Input *in, xgi::Output *out)
+          throw (xgi::exception::Exception);
+
+        std::vector<Calibration*> v_gemcal;
```
jsturdy: What is this? I.e., why does the `Calibration` class contain a vector of pointers to `Calibration` items?
Use smart pointers unless impossible due to compatibility with external libraries

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+
+        calType_t m_calType;
+
+        std::map<calType_t, std::map<std::string, uint32_t>> m_scanParams{
```
jsturdy: This needs documentation

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+               {"calPhase"  , 0},
+		 // TODO: need to take the list of te ARM dac from a file
+	       }}  
+        };
+
+	struct  scanParamsRadioSelector{
+	  //std::string name;
+	  
+	  std::string label;
+	  std::vector<std::string>  radio_options;
```
jsturdy: is `radio_` necessary?
the struct type already labels it such

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+	  
+	};
+	
+    
+	std::map<std::string, scanParamsRadioSelector> m_scanParamsRadioSelector{
```
jsturdy: Without seeing how these (the various `map` objects herein defined) are used, I would encode these in `enums`.
Documentation *might* help decide whether that is reasonable.
Additionally, that it is labeled as a radio selector, it strongly indicates that it is rather part of the UI (`CalibrationWeb`) rather than part of the core functionality, (`Calibration`)
If they are not converted to appropriate enums, they should almost definitely be `const` or `constexpr`

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+        /* virtual bool calibrationSequencer(toolbox::task::WorkLoop *wl); */
+
+      private:
+        /**
+         * @param classname is the class to check to see whether it is a GEMApplication inherited application
+         * @throws
+         */
+        bool isGEMApplication(const std::string& classname) const;
+        xdata::Integer m_shelfID;
+        log4cplus::Logger m_logger; //FIXME should be removed!
```
jsturdy: Then why isn't it?

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+        /* virtual bool calibrationAction(toolbox::task::WorkLoop *wl); */
+        /* virtual bool calibrationSequencer(toolbox::task::WorkLoop *wl); */
+
+      private:
+        /**
+         * @param classname is the class to check to see whether it is a GEMApplication inherited application
+         * @throws
+         */
+        bool isGEMApplication(const std::string& classname) const;
+        xdata::Integer m_shelfID;
```
jsturdy: what is the purpose of this?

gemcalibration/include/gem/calib/CalibrationWeb.h
```diff
@@ -0,0 +1,91 @@
+/** @file CalibrationWeb.h */
+
+#ifndef GEM_GEMCAL_CALIBRATIONWEB_H
```
jsturdy: header guards should match the namespace, i.e., `GEM_CALIB_` (also elsewhere)

gemcalibration/src/common/Calibration.cc
```diff
@@ -0,0 +1,226 @@
...
+      CMSGEMOS_DEBUG("gem::calib::Calibration:actionPerformed() ItemChangedEvent");
+    }
+    
+    // update monitoring variables
+     gem::base::GEMApplication::actionPerformed(event);
+}
+
+void gem::calib::Calibration::init()
+{
+  /*  v_daqmon.clear();
```
jsturdy: Another set of needed to be removed commented lines

gemcalibration/src/common/Calibration.cc
```diff
@@ -0,0 +1,226 @@
...
+
+void gem::calib::Calibration::pauseAction(xgi::Input* in, xgi::Output* out)
+  throw (xgi::exception::Exception)
+{
+  
+  CMSGEMOS_INFO("Calibration::pauseAction");
+  out->getHTTPResponseHeader().addHeader("Content-Type", "application/json");
+  //this->stopMonitoring();
+  *out << " { \"mon_state\":\"PAUSED\"}" << std::endl;
+  std::cout << "ciao! " << std::endl;
```
jsturdy: Use logger, not `cout`

gemcalibration/src/common/Calibration.cc
```diff
@@ -0,0 +1,226 @@
...
+    bool t_errorsOccured = false;
+    cgicc::Cgicc cgi(in);
+    for (auto it: m_scanParams.find(m_calType)->second){
+        it.second = cgi[it.first]->getIntegerValue();
+        CMSGEMOS_DEBUG("Calibration::applyAction : " << it.first << " = " << it.second);
+    }
+    
+    std::stringstream t_stream;
+    t_stream.clear();
+    t_stream.str(std::string());
```
jsturdy: Are these two lines necessary before the stream has ever been used?

setup/etc/profile.d/gemdaqenv.sh
jsturdy: This file was *not* listed in my instructions for cleaning up the old PR...

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+          {"scanMax"   , "Scan max"},
+          {"vfatChMin" , "VFAT Ch min"},
+          {"vfatChMax" , "VAT Ch max"},
+          {"vt2"       , "CFG_THR_ARM_DAC"},
+          {"trigThrottle"  , "Trigger throttle (int)"},
+	  {"pulseDelay", "Pulse delay (BX)"},
+	  {"latency"   , "Latency (BX)"},
+	  {"trimValues", "Points in dac range (odd)"}, // TODO:need to be implemented properly in the back end in order to get a given number of points {-63,0,63}
+	    };
+        std::map<std::string, uint32_t> amc_optical_links;
```
jsturdy: member variables must use `m_<name>` where `<name>` should utilize "almost" `CamelCase`/`PascalCase` (see updated `CONTRIBUITING.md` guidelines)

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,240 @@
...
+//#include "gem/daqmon/DaqMonitor.h"
+
+#define NAMC 12
+#define NSHELF 2
+
+
+namespace gem {
+  namespace calib {
+
+    enum calType {NDEF, GBTPHASE, LATENCY, SCURVE, SBITARMDACSCAN, ARMDACSCAN, TRIMDAC, DACSCANV3, CALIBRATEARMDAC};
```
jsturdy: I would put these `enums` in a dedicated file (e.g., `GEMCalibEnums.h`).
See, e.g., `gembase/include/gem/base/utils/GEMEnums.h`

jsturdy (changes requested):

bdorney:
Is it possible to get a pdf like last time that shows the final interface that this PR creates for each scan type?

cgalloni:
Hi Brian, yes,  I'll do that.

cgalloni:
Hi @bdorney, meanwhile I implement the other comments see here the current interface. 
[CalibrationInterface2.pdf](https://github.com/cms-gem-daq-project/cmsgemos/files/3108460/CalibrationInterface2.pdf)

config/mfPythonDefsGEM.mk
cgalloni: I was pointed to this https://github.com/cms-gem-daq-project/cmsgemos/blob/generic-amc-release-v3/config/mfPythonDefsGEM.mk because of problems during the compilation related to the file Python.h not found.

gemcalibration/include/gem/calib/Calibration.h
cgalloni: changed to simply `options`.

gemcalibration/include/gem/calib/Calibration.h
cgalloni: Radio selectors as part of the CalibrationWeb UI are implemented for those parameters that don't need forms, but just can assume certain values. For instance for the TriggerType, ComparatorType to be used, etc.. These parameters will need to be fed to the back-end code eventually.

gemcalibration/include/gem/calib/Calibration.h
cgalloni: removed

gemcalibration/include/gem/calib/CalibrationWeb.h
cgalloni: done

config/mfPythonDefsGEM.mk
mexanick: Actually this has to be changed to python compilation flags because that has been changed towards the develop...
https://github.com/cms-gem-daq-project/cmsgemos/blob/develop/config/mfPythonDefsGEM.mk

gemcalibration/include/gem/calib/Calibration.h
mexanick: looks like remnant from monitoring package... Indeed has to be removed

config/mfPythonDefsGEM.mk
jsturdy: This *should* be taken care of by the changes to the `gemcalibration/Makefile`, which was why I was curious why this change was introduced.

```
UserCFlags  +=$(PYTHONCFLAGS)
UserCCFlags +=$(PYTHONCFLAGS)
```

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,307 @@
...
+  throw (xgi::exception::Exception)
+{
+  *out << cgicc::script().set("type", "text/javascript")
+    .set("src", "/gemdaq/gemcalibration/html/scripts/cal_routines.js")
+       << cgicc::script() << std::endl;
+
+  *out << "<link  rel=\"stylesheet\" href=\"/gemdaq/gembase/html/css/bootstrap.css\" type=\"text/css\">" << std::endl;
+  *out << "<link  rel=\"stylesheet\" href=\"/gemdaq/gembase/html/css/bootstrap.min.css\" type=\"text/css\">" << std::endl;
+  *out << "<script src=\"/gemdaq/gembase/html/scripts/bootstrap.js\"></script>" << std::endl;
+  *out << "<script src=\"https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js\"></script>" << std::endl;
```
jsturdy: External links don't work on the technical network, please select a version and do as you have done with `bootstrap`

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,307 @@
...
+void gem::calib::CalibrationWeb::genericRadioSelector( std::string paramName, gem::calib::Calibration::scanParamsRadioSelector radio_param  ,xgi::Output* out)
+  throw (xgi::exception::Exception)
+{
+  *out << "<h2><span class=\"label label-warning col-md-6\">"<< radio_param.label << "</span></h2>"<< std::endl;
+  *out << "<div class=\"col-md-6\">"<< std::endl;
+  *out << "    <div class=\"form-check\">" << std::endl;
+
+  for (unsigned int i=0; i< radio_param.options.size();i++) {
+    if (i==0) {
+      *out<<"<div class=\"form-check-input radio-inline\"> <label><input type=\"radio\" name=\""<<paramName<<"\" id=\""<<paramName<<"_radio_"<<i<<"\" value="<<i<<" checked>"<<radio_param.options[i]<<"</label></div>"<< std::endl;
```
jsturdy: If you want to check your HTML indentation, you can load up your page in a browser and then view the source.
Here you've moved the indentation to the beginning of the HTML page, which is not likely what you wanted

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,307 @@
...
+
+
+void gem::calib::CalibrationWeb::genericRadioSelector( std::string paramName, gem::calib::Calibration::scanParamsRadioSelector radio_param  ,xgi::Output* out)
+  throw (xgi::exception::Exception)
+{
+  *out << "<h2><span class=\"label label-warning col-md-6\">"<< radio_param.label << "</span></h2>"<< std::endl;
+  *out << "<div class=\"col-md-6\">"<< std::endl;
+  *out << "    <div class=\"form-check\">" << std::endl;
+
+  for (unsigned int i=0; i< radio_param.options.size();i++) {
```
jsturdy: Prefer preincrement unless specifically need a postincremenet

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,307 @@
...
+  *out << "        </div>"  << std::endl;
+  *out << "    </div>" << std::endl; // end panel heaing
+  // panel body
+  *out << "    <form id=\"dacScanV3_select\">" << std::endl;
+  *out << "        <div class=\"container\" id=\"dacScanV3_selection\">" << std::endl;
+  *out << "            <div class=\"dropdown\">" << std::endl;
+  *out << "                <button id=\"dacScan_dropdown_button\" class=\"btn btn-lg btn-outline dropdown-toggle\" data-toggle=\"dropdown\">Settings</button>" << std::endl;
+  *out << "                <div class=\"dropdown-menu pre-scrollable\">" << std::endl;
+
+  for(auto it=dynamic_cast<gem::calib::Calibration*>(p_gemApp)->m_dacScanTypeParams.begin();it!=dynamic_cast<gem::calib::Calibration*>(p_gemApp)->m_dacScanTypeParams.end();it++){
```
jsturdy: You can probably clean this up with 
```c++
for (auto & it : dynamic_cast<gem::calib::Calibration*>(p_gemApp)->m_dacScanTypeParams) {
```

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,244 @@
...
+#include "gem/calib/GEMCalibEnums.h"
+
+
+namespace gem {
+  namespace calib {
+    
+    typedef enum calType calType_t;
+    
+    
+    typedef enum dacScanType dacScanType_t;
```
jsturdy: why not also put the typedefs in the enum defs header?

bdorney:
> Hi @bdorney, meanwhile I implement the other comments see here the current interface.
> [CalibrationInterface2.pdf](https://github.com/cms-gem-daq-project/cmsgemos/files/3108460/CalibrationInterface2.pdf)

Hi @cgalloni it seems like you addressed a lot of the comments we had the last time on the interface layout.  Thanks a lot.  Another round of comments here.  Apologies if there is some duplication.

What is `CalPulse phase` here supposed to be?  Is this the time between L1A and calpulse? Or is this the phase in the VFAT3 calibration unit? e.g. `GEM_AMC.OH.OHX.GEB.VFATY.CFG_CAL_PHI`?

Now slide specific comments.

# Slide 3

- A GBT Phase scan does not need a trigger source, drop this.
- Add the following additional fields:
   - phase min (int) defaulting to 0,
   - phase max (int) defaulting to **14** (we've seen phase 15 is always bad, so no reason to include it)
   - step size (int) defaulting to 1

Also the yellow bar says `HELLOW WORLD`.  I guess this is not intended?

# Slide 4

- Change default `ohMask` to `0xfff` since we should assume in P5 we will have 12 OH's per CTP7; or default it to `0xffc` since on QC8 `OH0` and `OH1` are not used per CTP7.  But the `0x3ff` is a legacy from the slice test.

# Slide 5

- Change default `Throttle` to `0`; our use case in near term is entirely cosmic data taking or internal cal pulse and for cosmic I fear having a default of 100 will cause the scans to take 100 times as longer :D
- Change default `CFG_THR_ARM_DAC` to at least 50 or 100.  Having a default of 0 will probably risk to end up recording only noise.

If the `calibration pulse` radial button is selected then the `Pulse Delay` field (e.g. from slide 6) should be present here.  It's meaningless for the `particle` option.

Here I assume `Pulse Delay` is the difference between the calpulse and the L1A and this should be configurable for a latency scan with the calibration pulse."

# Slide 6

- What's the rational in including a field for `CFG_THR_ARM_DAC` on slide 5 for the latency scan but not here?  Just wondering if we should be adding/dropping this; Is the use case the user configures with some known config in the DB? Or makes a config by hand and then makes a test?  Or...?  If a known config from the DB is to be used then this field should be dropped from slide 5 and not added here.
- The `Pulse stretch (int)` field is needed here.  Default to 4 is fine.

# Slide 7

- Drop field `Number of samples` it is meaningless here
- Add field `Step Size (int)` with default of 1.

# Slide 8

- Increase default number of samples to 1000

# Slide 9

- In previous slides were calibration module is used you have a calphase input field but not here, if we should allow this to be a variable parameter it should be here as well.  Could you add this?
- How does `Points in dac range (odd)` transform to the actual trim points that are being used?  e.g. if I enter 5 what trim DAC values are used?  Why not just have a comma separated list of trim DAC values defaulting to `-63,0,63`?  This to me is easier to understand.
- Also it was unclear to me what the final goal of *this* was.  Did we decide to have this only taking the scurves followed by some "offline" analysis to derive a config that is tested in a later scurve run?  Maybe more a question for @mexanick.

# Slide 10

- Right now the routine that performs this does not take an input for the number of ADC reads (e.g. number of samples) to perform for each DAC value (it's presently hard coded to 100).  So perhaps this field should be dropped entirely.

# Slide 11

- Nice drop down here, I would set the defaults such that all boxes were checked *except* the following:
   - `CFG_CAL_DAC`, and
   - `CFG_VREF_ADC`

This would represent the most useful default behavior.  You mentioned it's possible to easily select all, is it also easy to deselect all? If not could you add a `deselect all` button?

# Slide 12

Could you add a field that specifies the ARM DAC points to use for the calibration which defaults to the following list: `17,20,23,25,30,40,50,60,70,80,100,125,150,175`?

bdorney (changes requested):
Changes on interface layout requested.

jsturdy (changes requested):

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,244 @@
...
+	  {CFG_BIAS_SD_I_BFCAS, "CFG_BIAS_SD_I_BFCAS"},
+	  {CFG_BIAS_SD_I_BSF, "CFG_BIAS_SD_I_BSF"},
+	  {CFG_BIAS_CFD_DAC_1,"CFG_BIAS_CFD_DAC_1"},
+	  {CFG_BIAS_CFD_DAC_2,"CFG_BIAS_CFD_DAC_2"},
+	  {CFG_HYST,"CFG_HYST"},
+	  {CFG_THR_ARM_DAC,"CFG_THR_ARM_DAC"},
+	  {CFG_THR_ZCC_DAC, "CFG_THR_ZCC_DAC"},
+	  {CFG_BIAS_PRE_VREF,"CFG_BIAS_PRE_VREF"},
+	  {CFG_VREF_ADC, "CFG_VREF_ADC"},
+	};
```
jsturdy: * Can all these various maps be `const`/`constexpr`?
  * They seem to simply be placeholders for templates for the UI display defaults.
* Some of them seem to be duplicated unnecessarily:
  * `m_dacScanTypeParams_label` and `m_dacScanTypeParams` could be merged to give a mapping:
    * `enum` -> `{label, min, max}`

bdorney:
Another question; this interface appears to let a user set these values and then execute a scan. Suppose I want to come back to a long scan some time later when it’s still running. How can I check the entered values at runtime? Through RCMS? And how can I check what the current position in the scan is?

mexanick:
> Another question; this interface appears to let a user set these values and then execute a scan. Suppose I want to come back to a long scan some time later when it’s still running. How can I check the entered values at runtime? Through RCMS? 

This is a good question. So far we don't have a well-established mechanism. The easiest thing to do is to disable form controls (read buttons, input fields etc.) but continue showing the screen. However it would be better to store the scan configuration in the DB. But this has to come later, after we manage to integrate the DB interactions for detector configurations 

> And how can I check what the current position in the scan is?

My idea was to implement a progress bar based on the range and its execution. Will be added when the routines itself will be attached.

bdorney:
> This is a good question. So far we don't have a well-established mechanism. The easiest thing to do is to disable form controls (read buttons, input fields etc.) but continue showing the screen. However it would be better to store the scan configuration in the DB. But this has to come later, after we manage to integrate the DB interactions for detector configurations

Disabling the form controls would be nice; I assume the running xdaq process would show this, i.e. if another user navigates to the hyperdaq page they can see this info as well.

I agree it would be good to have scan configuration in the DB as well.

> My idea was to implement a progress bar based on the range and its execution. Will be added when the routines itself will be attached.

Great!

cgalloni:
Hi @bdorney,   please see the ansewrs in line.

> What is CalPulse phase here supposed to be? Is this the time between L1A and calpulse? Or is this the phase in the VFAT3 calibration unit? e.g. GEM_AMC.OH.OHX.GEB.VFATY.CFG_CAL_PHI?

This was intetended to be the pulse delay,  so it was changed to "Pulse delay" and set as default to 40

> Now slide specific comments.
> Slide 3
> 
>     A GBT Phase scan does not need a trigger source, drop this.
>     Add the following additional fields:
>         phase min (int) defaulting to 0,
>         phase max (int) defaulting to 14 (we've seen phase 15 is always bad, so no reason to include it)
>         step size (int) defaulting to 1
> Also the yellow bar says HELLOW WORLD. I guess this is not intended?

Changed

> Slide 4
> 
>     Change default ohMask to 0xfff since we should assume in P5 we will have 12 OH's per CTP7; or default it to 0xffc since on QC8 OH0 and OH1 are not used per CTP7. But the 0x3ff is a legacy from the slice test.

Changed to 0xffc for QC8.

> Slide 5
> 
>     Change default Throttle to 0; our use case in near term is entirely cosmic data taking or internal cal pulse and for cosmic I fear having a default of 100 will cause the scans to take 100 times as longer :D
>     Change default CFG_THR_ARM_DAC to at least 50 or 100. Having a default of 0 will probably risk to end up recording only noise.
> 
> If the calibration pulse radial button is selected then the Pulse Delay field (e.g. from slide 6) should be present here. It's meaningless for the particle option.
> 
> Here I assume Pulse Delay is the difference between the calpulse and the L1A and this should be configurable for a latency scan with the calibration pulse."
> 
> Slide 6
> 
>     What's the rational in including a field for CFG_THR_ARM_DAC on slide 5 for the latency scan but not here? Just wondering if we should be adding/dropping this; Is the use case the user configures with some known config in the DB? Or makes a config by hand and then makes a test? Or...? If a known config from the DB is to be used then this field should be dropped from slide 5 and not added here.
>     The Pulse stretch (int) field is needed here. Default to 4 is fine.

Changes implemented.  The CFG_THR_ARM_DAC  was masked form the interface as it should be loaded from DB. 


> Slide 9
> 
>     In previous slides were calibration module is used you have a calphase input field but not here, if we should allow this to be a variable parameter it should be here as well. Could you add this?

`Calphase` parameter was removed form the rest of the interface and here there is the Pulse Delay parameter.   


>    How does Points in dac range (odd) transform to the actual trim points that are being used? e.g. if I enter 5 what trim DAC values are used? Why not just have a comma separated list of trim DAC values defaulting to -63,0,63? This to me is easier to understand.
>     Also it was unclear to me what the final goal of this was. Did we decide to have this only taking the scurves followed by some "offline" analysis to derive a config that is tested in a later scurve run? Maybe more a question for @mexanick.

Implementd a string with {-63, 0, 63}.

> Slide 11
> 
>     Nice drop down here, I would set the defaults such that all boxes were checked except the following:
>         CFG_CAL_DAC, and
>         CFG_VREF_ADC
> 
> This would represent the most useful default behavior. You mentioned it's possible to easily select all, is it also easy to deselect all? If not could you add a deselect all button?

Yes, at the moment there are 3 buttons " DEFAULT" ,"SELECT  ALL" and "DESELECT ALL"

Slide 12

Could you add a field that specifies the ARM DAC points to use for the calibration which defaults to the following list: 17,20,23,25,30,40,50,60,70,80,100,125,150,175?

Done. I'll prepare a set of slides with the current layout.

cgalloni:
Hi @bdorney @jsturdy  @mexanick,
here is the new version of the calibration interface.  Please let me know if you have further comments. 
Thanks and cheers,

[CalibrationInterface3.pdf](https://github.com/cms-gem-daq-project/cmsgemos/files/3127515/CalibrationInterface3.pdf)

bdorney:
> CalibrationInterface3.pdf

I think slide 5 has the GBT scan interface instead of the latency scan interface by mistake?

bdorney:
> Hi @bdorney @jsturdy @mexanick,
> here is the new version of the calibration interface. Please let me know if you have further comments.
> Thanks and cheers,
> 
> [CalibrationInterface3.pdf](https://github.com/cms-gem-daq-project/cmsgemos/files/3127515/CalibrationInterface3.pdf)

So for every slide except slide 5 (latency scan) the interface looks great from my side.  No further comments there.  If you can re-upload a version that has slide 5 for latency scan I'll take a look at that later.

cgalloni:
Hi @bdorney , 
yes, you are right , I made a mistake with the cut and paste. Apologies.
Here the correct one:
[CalibrationInterface3.pdf](https://github.com/cms-gem-daq-project/cmsgemos/files/3128158/CalibrationInterface3.pdf)

bdorney:
> Hi @bdorney ,
> yes, you are right , I made a mistake with the cut and paste. Apologies.
> Here the correct one:
> [CalibrationInterface3.pdf](https://github.com/cms-gem-daq-project/cmsgemos/files/3128158/CalibrationInterface3.pdf)

Thanks looks good. No objections from my side.

bdorney (dismissed):

mexanick (dismissed):
Looks good to me

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,224 @@
...
+            };
+
+            std::map<std::string, uint32_t> m_amcOpticalLinks;
+
+            dacScanType_t m_dacScanType;
+
+            struct dacScanTypeFeature{
+                std::string label;
+                std::map<std::string, uint32_t> range;
+            };
```
jsturdy, line 167: OK, let me be more explicit.
For the usage as I've understood it here, the type that would be most appropriate would be a `union`
```c++
union dacFeature {
   std::string label;
   uint16_t min;
   uint16_t max;
};
```

gemcalibration/include/gem/calib/Calibration.h
```diff
@@ -0,0 +1,224 @@
...
+        private:
+            /**
+             * @param classname is the class to check to see whether it is a GEMApplication inherited application
+             * @throws
+             */
+            bool isGEMApplication(const std::string& classname) const;
+
+            xdata::Integer m_nShelves;
+
+            log4cplus::Logger m_logger;
```
jsturdy: In a previous iteration, this line had a comment to be removed, is that not the case?

gemcalibration/include/gem/calib/CalibrationWeb.h
```diff
@@ -0,0 +1,135 @@
...
+                };
+            // GEMSupervisor *gemSupervisorP__;
+            // GEMSupervisorWeb(GEMSupervisorWeb const&);
+        };  // class gem::calib::CalibrationWeb
+    }  // namespace gem::calib
+}  // namespace gem
+
+#endif  // GEM_CALIB_CALIBRATIONWEB_H
+
+// Local Variables:
```
jsturdy: This type of stuff is generally frowned upon in this repository, please remove

gemcalibration/src/common/Calibration.cc
```diff
@@ -0,0 +1,172 @@
...
+    gem::base::GEMApplication::actionPerformed(event);
+}
+
+void gem::calib::Calibration::init()
+{
+    CMSGEMOS_INFO("gem::calib::Calibration:init() number of shelves: " << m_nShelves.value_);
+ 
+}
+
+bool gem::calib::Calibration::isGEMApplication(const std::string& classname) const
```
jsturdy: I don't think this function is relevant for this application

gemcalibration/src/common/CalibrationWeb.cc
```diff
@@ -0,0 +1,307 @@
...
+    *out << "        <div class=\"col-md-6\">" << std::endl;
+    *out << "            <form id=\"settings_select\">" << std::endl;
+
+    // create a temporary parameters map,
+    // check whether requested calibration type needs a trigger selector
+    // if it is, create it and pop out the trigger setting
+    std::map<std::string, std::string> t_parameters = dynamic_cast<gem::calib::Calibration*>(p_gemApp)->m_scanParams.find(m_calType)->second; 
+
+    for (auto parameter: t_parameters) {
+        if (dynamic_cast<gem::calib::Calibration*>(p_gemApp)->m_scanParamsNonForm.find(parameter.first) !=dynamic_cast<gem::calib::Calibration*>(p_gemApp)->m_scanParamsNonForm.end()) {
```
jsturdy: As you use this conversion multiple times in this function (and in other places where you are doing multiple casts), probably better to create a local variable, e.g.,
```c++
auto gemAppLocal = dynamic_cast<gem::calib::Calibration*>(p_gemApp);
```

jsturdy (changes requested):

gemcalibration/include/gem/calib/Calibration.h
jsturdy: As @cgalloni found, in fact `union` was *not* the appropriate type for this usage, but a `struct` with two additional fields rather than a `map` was.

jsturdy (changes requested):
Can you squash your last two commits into one (so the removal of commented lines, and the structural change that introduced those commented lines) are a single commit?

mexanick:
@jsturdy I guess all your comments has been addressed... Please check this

jsturdy (approved):
> @jsturdy I guess all your comments has been addressed... Please check this

Yes, all other issues I have with the code can be addressed in future improvements

PR #276, opened 2019-04-25
OnlineDB monitoring page

lmoureaux:
## Description

Fill the "Monitoring" tab of the GEMWebApplication with some high-level information about the currently loaded configuration.

### Types of changes

## Motivation and Context

Allows one to easily check which files the configuration was pulled from.

## How Has This Been Tested?

Used the example data to produce the screenshot below.

### Screenshot:

Screenshot_20190425_143731

## Checklist:

- [ ] My code follows the code style of this project (but it's consistent with the style of the modified files).

### DIFF
--- a/gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
+++ b/gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
@@ -4,10 +4,38 @@
 
 #include <memory>
 
+#include <jansson.h>
+
 #include "xcept/tools.h"
 
+#include "gem/onlinedb/ConfigurationManager.h"
 #include "gem/onlinedb/GEMOnlineDBManager.h"
 #include "gem/onlinedb/exception/Exception.h"
+
+namespace /* anonymous */
+{
+    /// @brief Turns Jansson error codes into @c xgi::exception::Exception.
+    void janssonCheck(int retcode)
+    {
+        if (retcode != 0) {
+            XCEPT_RAISE(xgi::exception::Exception,
+                        "Call to Jansson function failed");
+        }
+    }
+    /**
+     * @brief Checks that the provided @c json_t is not null.
+     * @returns The provided pointer if it isn't null.
+     * @throws xgi::exception::Exception if the pointer is null.
+     */
+    json_t *janssonCheckPtr(json_t *json)
+    {
+        if (json == nullptr) {
+            XCEPT_RAISE(xgi::exception::Exception,
+                        "Jansson could not allocate memory");
+        }
+        return json;
+    }
+} // anonymous namespace
 
 gem::onlinedb::GEMOnlineDBManagerWeb::GEMOnlineDBManagerWeb(gem::onlinedb::GEMOnlineDBManager* dbApp) :
   gem::base::GEMWebApplication(dbApp)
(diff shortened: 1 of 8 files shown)


gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
```diff
@@ -65,9 +152,87 @@ void gem::onlinedb::GEMOnlineDBManagerWeb::applicationPage(xgi::Input* in, xgi::
...
+  // We use jsonUpdate for the basic statistics that are always visible in the
+  // first tab. Additional information is retrieved from other sources.
+
   CMSGEMOS_DEBUG("GEMOnlineDBManagerWeb::jsonUpdate");
   out->getHTTPResponseHeader().addHeader("Content-Type", "application/json");
-  *out << " { " << std::endl;
-  // fill JSON if needed
-  *out << " } " << std::endl;
+
+  json_auto_t *answer = janssonCheckPtr(json_object());
```
jsturdy, line 161: Ugg, is there a way to use smart pointers for these (and following to L237)?

gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
```diff
@@ -4,11 +4,39 @@
...
+            XCEPT_RAISE(xgi::exception::Exception,
+                        "Call to Jansson function failed");
+        }
+    }
+    /**
+     * @brief Checks that the provided @c json_t is not null.
+     * @returns The provided pointer if it isn't null.
+     * @throws xgi::exception::Exception if the pointer is null.
+     */
+    json_t *janssonCheckPtr(json_t *json)
```
jsturdy, line 30: Also here (re possibility of smart pointers)

jsturdy (commented):
Just a couple questions about memory

gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
lmoureaux: Jansson objects are reference-counted, so standard C++ smart pointers won't work (notice there's no call to `free()`). `json_auto_t` is a smart pointer that decrements the reference count when going out of scope (using `__attribute__((cleanup(function)))`).

gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
lmoureaux: I can't rely on `operator new` throwing `std::bad_alloc`, so I need to perform the error checking C-style. I don't see how smart pointers could improve the situation.

lmoureaux:
In xDAQ 15 we'll have a modern [C++ library for JSON](https://github.com/nlohmann/json) (thanks @lpetre-ulb for letting me know). We could use the version from EPEL (`json-devel`) in the meantime.

gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
jsturdy: OK, I was mostly concerned because the last line I mention explicitly calls `free` on the `json` object, so I was concerned.
Maybe my comment should only apply to L235-237.
If you're not leaking, then I guess I won't object.

gemonlinedb/src/common/GEMOnlineDBManagerWeb.cc
```diff
@@ -65,9 +152,87 @@ void gem::onlinedb::GEMOnlineDBManagerWeb::applicationPage(xgi::Input* in, xgi::
...
+        "vfat-count",
+        janssonCheckPtr(json_integer(stats.vfatCount))));
+      janssonCheck(json_object_set(answer, "statistics", statsJson));
+    }
+  }
+
+  // Write to stream
+  char *encoded = json_dumps(answer, 0);
+  *out << encoded;
+  free(encoded);
```
lmoureaux, line 237: This `free` could indeed be avoided using a smart pointer, but it may be overkill.

lmoureaux:
Rebased on top of current `develop`.

jsturdy (commented):
@lmoureaux, if you confirm that this isn't leaking, I'll sign off

lmoureaux:
Running `valgrind --leak-check=full --show-leak-kinds=definite xdaq.exe ...` and doing exactly one request to `updateJson`:

## Existing leaks

```
==11425== LEAK SUMMARY:
==11425==    definitely lost: 310 bytes in 4 blocks
==11425==    indirectly lost: 31 bytes in 1 blocks
==11425==      possibly lost: 160,854,861 bytes in 3,138 blocks
==11425==    still reachable: 10,726,585 bytes in 1,958 blocks
==11425==         suppressed: 0 bytes in 0 blocks
==11425== Reachable blocks (those to which a pointer was found) are not shown.
```

Largest leak is in xDAQ:
```
==8763== 150 bytes in 1 blocks are definitely lost in loss record 1,709 of 2,002
==8763==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==8763==    by 0x7E3F640: __libc_alloc_buffer_allocate (alloc_buffer_allocate.c:26)
==8763==    by 0x7EE9097: alloc_buffer_allocate (alloc_buffer.h:143)
==8763==    by 0x7EE9097: __resolv_conf_allocate (resolv_conf.c:410)
==8763==    by 0x7EE6093: __resolv_conf_load (res_init.c:587)
==8763==    by 0x7EE85A8: __resolv_conf_get_current (resolv_conf.c:162)
==8763==    by 0x7EE6DFC: __res_vinit (res_init.c:609)
==8763==    by 0x7EE7F9F: maybe_init (resolv_context.c:122)
==8763==    by 0x7EE7F9F: context_get (resolv_context.c:184)
==8763==    by 0x7EE7F9F: __resolv_context_get (resolv_context.c:195)
==8763==    by 0x7ED781B: gethostbyname (getXXbyYY.c:105)
==8763==    by 0x595C645: toolbox::net::getHostName[abi:cxx11]() (Utils.cc:34)
==8763==    by 0x4E78653: xdaq::ApplicationContextImpl::init(int, char**) (ApplicationContextImpl.cc:287)
==8763==    by 0x10CF3C: main (stl_tree.h:210)
```
The leak is at `toolbox/src/common/net/Utils.cc:34` (obvious when looking at the source code). How can I report issues against xDAQ?

## After changes to `jsonUpdate`

No new leaks detected.

jsturdy (approved):
OK, thanks for the check!

> The leak is at `toolbox/src/common/net/Utils.cc:34` (obvious when looking at the source code). How can I report issues against xDAQ?

You can open a ticket on their [JIRA](https://its.cern.ch/jira/projects/CMSOS/issues/?filter=allopenissues)

lmoureaux:
Checked the "xDAQ" leak again. The return value of `gethostbyname` should not be freed as I assumed, as it will be reused between different calls. So it's not a bug but a nice feature of this (obsolete) function.

PR #282, opened 2019-06-04
Updated assignment of number of GBTs for different gemTypes

sbutalla:
## Description
Added a logic statement to correctly assign the number of GBTs for the appropriate gemType (i.e., ge21 has two GBTs and ge11 has three GBTs).

### Types of changes

## Motivation and Context
Number of GBTs was being assigned incorrectly to the ge21 OH as three instead of two.

## Checklist:

### DIFF
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -75,7 +75,7 @@ class NoUnmaskedOHException(Exception):
         return
 
 class HwAMC(object):
-    def __init__(self, cardName, debug=False, args.gemType="ge11"):
+    def __init__(self, cardName, debug=False, gemType="ge11"):
         """
         Initialize the HW board an open an RPC connection
         """
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -75,7 +75,7 @@ class NoUnmaskedOHException(Exception):
         return
 
 class HwAMC(object):
-    def __init__(self, cardName, debug=False):
+    def __init__(self, cardName, debug=False, args.gemType="ge11"):
         """
         Initialize the HW board an open an RPC connection
         """
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -185,7 +185,6 @@ class HwAMC(object):
         """
         Using the SBIT_MONITOR acquire sbit data for time in seconds given
         by acquireTime and write the data to the directory specified by outFilePath
-
         ohN - optohybrid to monitor
         outFilePath - filepath created data files will be written too
         acquireTime - time in seconds to acquire data for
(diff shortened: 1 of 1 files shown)


gempython/tools/amc_user_functions_xhal.py
```diff
@@ -279,7 +274,15 @@ def getGBTLinkStatus(self,doReset=False,printSummary=False, ohMask=None):
...
-        NGBT = gbtsPerGemVariant["ge11"]
+        #Determine the number of GBTs based on the gemType
+        if gemType == "ge11":
+            NGBT = gbtsPerGemVariant["ge11"]
+        elif gemType == "ge21":
+            NGBT = gbtsPerGemVariant["ge21"]
+        else:
+            NGBT = 0
+            print("Unrecognized gemType, setting NGBT = 0")
+
```
bdorney: Would strongly prefer the following, and in the `HwAMC::__init__(...)` method:

```python
if gemType in gbtsPerGemVariant.keys():
    self.NGBT = gbtsPerGemVariant[gemType]
else:
    raise KeyError("Unrecognized gemType {0}".format(gemType))
```

Then here you just use `self.NGBT`

bdorney (dismissed):
changes should go in `init` method and raise an exception if unrecognized key rather than just continuing.

gempython/tools/amc_user_functions_xhal.py
bdorney: An alternative, but also acceptable implementation in `HwAMC::_init__(...), could also be:

```python
try:
   self.NGBT = gbtsPerGemVariant[gemType]
except KeyError as error:
   printRed("Unrecognized gemType {0}".format(gemType))
   exit(os.EX_DATAERR)
```

You might need to add `import os` and `from gempython.utils.gemLogger import printRed`.  I believe `printRed` is defined under `gempython/utils/gemLogger.py` but double check.

sbutalla: I agree, I will implement and resubmit the pull request.

gempython/tools/amc_user_functions_xhal.py
```diff
@@ -179,6 +179,12 @@ def __init__(self, cardName, debug=False):
         if debug:
             print "My FW release major = ", self.fwVersion
 
+        #Determine the number of GBTs based on the gemType
```
mexanick, line 182: I don't see how the gemType is passed to the constructor of the `HwAMC` class. Please update the `init` signature

mexanick (changes requested):

gempython/tools/amc_user_functions_xhal.py
sbutalla: Fixed! Thank you for spotting that.

mexanick (approved):

mexanick (changes requested):
the gemType should be declared as local variable, not as `args.gemType`. 
I.e.: 
```python 
def __init__(self, cardName, debug=False, gemType="ge11"):
```

mexanick (approved):

mexanick (approved):

bdorney (commented):
Either @sbutalla or a maintainer should `squash` the commit history to a single commit with a meaningful commit message.

mexanick:
my bad, forgot to squash

PR #287, opened 2019-06-13
Feature/ge21compatibility

mexanick:
## Description

Implements GE2/1 compatibility. All new parameters are backed with default values set to GE1/1 defaults. Thus the rest of the software should be able to work unchanged. Requires https://github.com/cms-gem-daq-project/ctp7_modules/pull/132 to work with GE2/1
### Types of changes

## Motivation and Context

Need to be able to use the current legacy software for GE2/1 integration stand
## How Has This Been Tested?

Tested with GE2/1 integration stand.
## Checklist:

### DIFF
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -125,11 +125,11 @@ class HwAMC(object):
         self.confDacMonitorMulti.restype = c_uint
 
         self.readADCsMulti = self.lib.readVFAT3ADCMultiLink
-        self.readADCsMulti.argTypes = [ c_uint, POINTER(c_uint32), POINTER(c_uint), c_bool ]
+        self.readADCsMulti.argTypes = [ c_uint, POINTER(c_uint32), POINTER(c_uint), c_bool, c_uint ]
         self.readADCsMulti.restype = c_uint
 
         self.dacScanMulti = self.lib.dacScanMultiLink
-        self.dacScanMulti.argTypes = [ c_uint, c_uint, c_uint, c_uint, c_bool, POINTER(c_uint) ]
+        self.dacScanMulti.argTypes = [ c_uint, c_uint, c_uint, c_uint, c_bool, POINTER(c_uint), c_uint ]
         self.dacScanMulti.restype = c_uint
 
         # Define SCA & Sysmon Monitoring
@@ -165,7 +165,7 @@ class HwAMC(object):
         self.readSBits.restype = c_uint
 
         self.sbitRateScanMulti = self.lib.sbitRateScan
-        self.sbitRateScanMulti.argTypes = [c_uint, c_uint, c_uint, c_uint, c_uint, c_char_p, POINTER(c_uint32), POINTER(c_uint32), POINTER(c_uint32)]
+        self.sbitRateScanMulti.argTypes = [c_uint, c_uint, c_uint, c_uint, c_uint, c_char_p, POINTER(c_uint32), POINTER(c_uint32), POINTER(c_uint32), c_uint]
         self.sbitRateScanMulti.restype = c_uint
 
         # Parse XML
@@ -668,7 +668,7 @@ class HwAMC(object):
             printRed("HwAMC::performDacScanMultiLink(): I expected container of length {0} but provided 'dacDataAll' has length {1}",format(lenExpected, len(dacDataAll)))
             exit(os.EX_USAGE)
 
-        return self.dacScanMulti(ohMask, self.nOHs, dacSelect, dacStep, useExtRefADC, dacDataAll)
+        return self.dacScanMulti(ohMask, self.nOHs, dacSelect, dacStep, useExtRefADC, dacDataAll, self.NVFAT)
 
     def performSBITRateScanMultiLink(self, outDataDacVal, outDataTrigRate, outDataTrigRatePerVFAT, chan=128, dacMin=0, dacMax=254, dacStep=1, ohMask=None, scanReg="THR_ARM_DAC"):
         """
@@ -724,7 +724,7 @@ class HwAMC(object):
             printRed("HwAMC::performSBITRateScanMultiLink(): I expected container of length {0} but provided 'outDataTrigRatePerVFAT' has length {1}".format(self.NVFAT*lenExpected, len(outDataTrigRatePerVFAT)))
             exit(os.EX_USAGE)
 
-        return self.sbitRateScanMulti(ohMask, dacMin, dacMax, dacStep, chan, scanReg, outDataDacVal, outDataTrigRate, outDataTrigRatePerVFAT)
+        return self.sbitRateScanMulti(ohMask, dacMin, dacMax, dacStep, chan, scanReg, outDataDacVal, outDataTrigRate, outDataTrigRatePerVFAT, self.NVFAT)
 
     def programAllOptohybridFPGAs(self, maxIter=5, ohMask=None):
         """
@@ -828,7 +828,7 @@ class HwAMC(object):
             for ohN in range(0,12):
                 print("| {0} | 0x{1:x} |".format(ohN, ohVFATMaskArray[ohN]))
 
-        return self.readADCsMulti(ohMask,ohVFATMaskArray, adcDataAll, useExtRefADC)
+        return self.readADCsMulti(ohMask,ohVFATMaskArray, adcDataAll, useExtRefADC, self.NVFAT)
 
     def readBlock(register, nwords, debug=False):
         """
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -1,5 +1,5 @@
 from ctypes import *
-from gempython.tools.hw_constants import maxVfat3DACSize,gbtsPerGemVariant
+from gempython.tools.hw_constants import maxVfat3DACSize, gbtsPerGemVariant, vfatsPerGemVariant
 from gempython.utils.gemlogger import colors, printRed, printYellow
 from gempython.utils.wrappers import runCommand, runCommandWithOutput
 
@@ -182,6 +182,12 @@ class HwAMC(object):
         #Determine the number of GBTs based on the gemType
         if gemType in gbtsPerGemVariant.keys():
             self.NGBT = gbtsPerGemVariant[gemType]
+        else:
+            raise KeyError("Unrecognized gemType {0}".format(gemType))
+
+        #Determine the number of VFATs per geb based on the gemType
+        if gemType in vfatsPerGemVariant.keys():
+            self.NVFAT = vfatsPerGemVariant[gemType]
         else:
             raise KeyError("Unrecognized gemType {0}".format(gemType))
 
@@ -604,7 +610,7 @@ class HwAMC(object):
                 # print("----------OH{0}----------".format(ohN))
                 pass
 
-            for vfatN in range(24):
+            for vfatN in range(self.NVFAT):
                 nSyncErrors = vfatMonData[ohN].syncErrCnt[vfatN]
                 totalSyncErrors += nSyncErrors
 
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -807,7 +807,7 @@ class HwAMC(object):
     def readADCsMultiLink(self, adcDataAll, useExtRefADC=False, ohMask=None, debug=False):
         """
         Reads the ADC value from all unmasked VFATs
-        adcDataAll - Array of type c_uint32 of size self.NVFAT*12=288
+        adcDataAll - Array of type c_uint32 of size self.NVFAT*12 (288 for GE1/1 or 144 for GE2/1)
         useExtRefADC - True (False) use the externally (internally) referenced ADC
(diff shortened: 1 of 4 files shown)


jsturdy (commented):
please rebase

bdorney:
In the CTP7 modules PR we added things like `NUM_OH_PER_AMC`.  This is covered in `HwAMC` by a read of the register that says how many OH's are in the FW but one thing that's potentially overlooked are these lines:

https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L409-L411

e.g. the `0xfff` bit mask used in this operation is specific to 12 OH case...should this also be protected by a per `gemType` constant?

bdorney:
> In the CTP7 modules PR we added things like `NUM_OH_PER_AMC`. This is covered in `HwAMC` by a read of the register that says how many OH's are in the FW but one thing that's potentially overlooked are these lines:
> 
> https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L409-L411
> 
> e.g. the `0xfff` bit mask used in this operation is specific to 12 OH case...should this also be protected by a per `gemType` constant?

Same question here:

https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L464-L470

bdorney:
I think you need to add `self.NVFATs` here:

https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L665-L669

Specifically:

```python
        lenExpected = self.nOHs * (maxVfat3DACSize[dacSelect][0] - 0+1)*24 / dacStep
```

Also the following locations:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L723
- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L724

Also need `self.parentAMC.NVFATs` here:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/optohybrid_user_functions_xhal.py#L486

bdorney:
Need `self.NOHs` in following locations:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L828

bdorney:
Need documentation update in following locations I think:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L677-L693
- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L807-L814

bdorney:
Probably change how `mask` in broadcast methods works:

https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/optohybrid_user_functions_xhal.py#L106-L141

Propose to add to `HwOptohybrid`:

- `self.defaultVFATMask` as a per `gemType` parameter in the `__init__()` method
- Broadcast methods then have `mask` argument default to `None`
- If `mask` argument is `None` in broadcast methods the method will use `self.defaultVFATMask` which is the `gemType` specific usage; otherwise it uses the input `mask` argument.

Otherwise I think you'll run into bus errors at some point.

If above solution is adopted then the following method would also need a similar modification:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/vfat_user_functions_xhal.py#L346-L347

bdorney:
Looks like no changes to this method, e.g. documentation illustrates that `outData` should be of a size that expects 24 VFATs and 12 OHs.  Do we want to have unused memory like this?

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/optohybrid_user_functions_xhal.py#L143-L176

Same question for the below:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/optohybrid_user_functions_xhal.py#L350-L392

bdorney:
Probably need a per `gemType` constant for the number of channels in a detector, e.g.:

- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/vfat_user_functions_xhal.py#L135-L141
- https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/vfat_user_functions_xhal.py#L232-L243

bdorney (changes requested):
Outstanding questions and/or clarification needed.  Some additional compatibility issues raised.

bdorney:
I'm not sure I caught everything...

mexanick:
> > In the CTP7 modules PR we added things like `NUM_OH_PER_AMC`. This is covered in `HwAMC` by a read of the register that says how many OH's are in the FW but one thing that's potentially overlooked are these lines:
> > https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L409-L411
> > e.g. the `0xfff` bit mask used in this operation is specific to 12 OH case...should this also be protected by a per `gemType` constant?
> 
> Same question here:
> 
> https://github.com/mexanick/cmsgemos/blob/4cc3741fae8358643c885443a674c694686775ec/gempython/tools/amc_user_functions_xhal.py#L464-L470

I don't think a protections is needed here

gempython/tools/amc_user_functions_xhal.py
```diff
@@ -800,7 +807,7 @@ def programAllOptohybridFPGAs(self, maxIter=5, ohMask=None):
     def readADCsMultiLink(self, adcDataAll, useExtRefADC=False, ohMask=None, debug=False):
         """
         Reads the ADC value from all unmasked VFATs
-        adcDataAll - Array of type c_uint32 of size 24*12=288
+        adcDataAll - Array of type c_uint32 of size self.NVFAT*12=288
```
jsturdy: this can't be correct still, since you've added a potentially different constant value to the equation

jsturdy (changes requested):
One doc update

gempython/tools/amc_user_functions_xhal.py
mexanick: Oups, oversight =) Thanks!

gempython/tools/optohybrid_user_functions_xhal.py
```diff
@@ -163,7 +163,7 @@ def checkSbitMappingWithCalPulse(self, **kwargs):
         L1AInterval - Number of BX's inbetween L1A's
         mask        - VFAT mask to use for excluding vfats from the trigger
         nevts       - Number of events for each dac value in scan
-        outData     - Pointer to an array of size (24*128*8*nevts) which
+        outData     - Pointer to an array of size (self.parentAMC.nVFAT*128*8*nevts) which
```
bdorney: small issue, `parentAMC` does not have a data member (nor should it) called `nVFAT`.  I think there you mean `vfatsPerGemVariant[self.gemType]`?

gempython/tools/optohybrid_user_functions_xhal.py
```diff
@@ -334,15 +334,15 @@ def getTriggerSource(self):
...
+
     def getType(self):
         return (self.typeGEM,self.typeDet)
 
     def getVFATMask(self):
         """
         V3 electronics only
 
-        Returns a 24 bit number that should be used as the VFAT Mask
+        Returns a self.parentAMC.nVFAT bit number that should be used as the VFAT Mask
```
bdorney: [same](https://github.com/cms-gem-daq-project/cmsgemos/pull/287/files#r297221217)

gempython/tools/optohybrid_user_functions_xhal.py
```diff
@@ -370,7 +370,7 @@ def performCalibrationScan(self, **kwargs):
         mask        - VFAT mask to use
         nevts       - Number of events for each dac value in scan
         outData     - Array of type c_uint32, if chan >= 0 array size
-                      must be: ((dacMax - dacMin + 1) / stepSize) * 24.
+                      must be: ((dacMax - dacMin + 1) / stepSize) * self.parentAMC.nVFAT.
```
bdorney: [same](https://github.com/cms-gem-daq-project/cmsgemos/pull/287/files#r297221217)...apply this comment everywhere.

gempython/tools/amc_user_functions_xhal.py
```diff
@@ -185,6 +185,12 @@ def __init__(self, cardName, debug=False, gemType="ge11"):
         else:
             raise KeyError("Unrecognized gemType {0}".format(gemType))
 
+        #Determine the number of VFATs per geb based on the gemType
+        if gemType in vfatsPerGemVariant.keys():
+            self.NVFAT = vfatsPerGemVariant[gemType]
```
bdorney: we won't have a situation where `detType` has a different number of VFATs correct?
e.g. {M1,...,M8} all have 12 VFATs?

bdorney: Also in `HwOptohybrid` this is `self.nVFATs` vs. `self.NVFAT`.

Should we keep both these data members...in retrospect probably the better way to go would have been to have:

`HwOptohybrid` inherit from `HwAMC` and `HwVFAT` inherit from `HwOptohybrid`.  Then all this `parentX` crap which is, admittedly, messy could disappear.  Although whether that type of change is appropriate for this PR (or ever given this is #legacy) is unclear.  Alas...my programming skills have improved since the original design...

gempython/tools/optohybrid_user_functions_xhal.py
bdorney: Typo, should be `parentAMC.NVFAT` right? e.g. capital `N`?

gempython/tools/optohybrid_user_functions_xhal.py
bdorney: [same comment to be applied everywhere](https://github.com/cms-gem-daq-project/cmsgemos/pull/287/files#r297225193)

gempython/tools/optohybrid_user_functions_xhal.py
```diff
@@ -483,12 +483,12 @@ def performDacScan(self, outData, dacSelect, dacStep=1, mask=0x0, useExtRefADC=F
             exit(os.EX_USAGE)
 
         #Check length of results container
-        lenExpected = (maxVfat3DACSize[dacSelect][0] - 0+1)*24 / dacStep
+        lenExpected = (maxVfat3DACSize[dacSelect][0] - 0+1)*self.parentAMC.nVFAT / dacStep
```
bdorney: this will actually cause an `AttributeError` no? e.g. [same](https://github.com/cms-gem-daq-project/cmsgemos/pull/287/files#r297225193)

bdorney (changes requested):
some typos in function documentation and one spot where I think an `AttributeError` will be raised at runtime.

gempython/tools/amc_user_functions_xhal.py
jsturdy: I would caution against making API changes at this point, since this legacy stuff is supposed to be frozen anyway...
We will have to potentially think of how to map this in the updated API, where currently there is not going to *be* a `HwVFAT`, though it could be added if deemed absolutely critical (would just have to think of how best to do so...)

bdorney: > I would caution against making API changes at this point, since this legacy stuff is supposed to be frozen anyway...

Indeed.  For this PR then I would suggest we keep both data members but just  standardized the data member name in both classes for QOL; so it should be either `self.NVFAT` or `self.nVFATs` with a slight preference for `self.nVFATs` simply because it pre-dates this PR in `HwOptohybrid`.

jsturdy:
If @bdorney is satisfied with changes, please `rebase` to clean up the commits

jsturdy:
If @bdorney is satisfied with changes, please `rebase` to clean up the commits (not before as it will make the review more difficult)

bdorney:
> If @bdorney is satisfied with changes, please `rebase` to clean up the commits (not before as it will make the review more difficult)

My only outstanding comment was [here](https://github.com/cms-gem-daq-project/cmsgemos/pull/287#discussion_r297807012).

mexanick:
I believe all the comments has been addressed now

jsturdy (approved):
OK from my side

bdorney (approved):

jsturdy:
@mexanick, merging this *wouldn't* break backwards compatibility (GE1/1) until cms-gem-daq-project/xhal#119 and cms-gem-daq-project/ctp7_modules#132 are merged, right?

mexanick:
It should not break GE1/1 compatibility at all

jsturdy:
Grrr, @mexanick you merged... (I was going to squash)

mexanick:
ups. We can rewind if you insist

PR #288, opened 2019-06-15
[feature] HW device migration

jsturdy:
**BREAKING CHANGE!!**

## Description
Modifies `GEMHwDevice` to inherit from both `uhal::HwInterface` and `xhal::XHALInterface`

* `HwGenericAMC` and `HwOptoHybrid` have been updated, but possibly incompletely.
  * `gempython` library contains `boost::python` exports from HW device classes
* `GLIBManager` and `OptoHybridManager` have been minimally updated

Several legacy blocks are not removed yet, as they need a replacement or better restructuring.

This PR is *not* designed to be the final word on migration to a new framework, as we will need to adapt for the templated `RPC` methods.  But this PR should ensure that all legacy only functionality has been removed and all core functionality has been implemented.

### Types of changes

## Motivation and Context
* Needs correct handling of `xhal` devices (read: `RPC` connections)
  * Possibly `XHALInterface` should have a singelton as the RPC connection to a given endpoint, such that we don't create an RPC connection per `GEMHwDevice` that connects to the same hardware

This is targeting a `1.2.X` release scheme, but this is because from `2.X.Y` we will be strictly enforcing semver versioning, but need some time to get things into a stable enough state where that will be feasible.

Hopefully 2.X.Y will come relatively shortly after integrating this + `gemcalibration` + `gemonlinedb`

## How Has This Been Tested?
**UPDATE - 17-06-2019**
Testing of the `gempython` code on actual hardware has revealed a few tweaks that need to be made to make sure that the info scripts work out of the box.

##### `HwOptoHybrid`
```
T HwOptoHybrid::resetT1Counters() -> resetTTCCounters()
T HwOptoHybrid::updateT1Counters() -> updateTTCCounters()
T HwOptoHybrid::getT1Count(unsigned char const&, unsigned char const&) -> getTTCCount
T HwOptoHybrid::resetT1Count(unsigned char const&, unsigned char const&) -> resetTTCCount
T HwOptoHybrid::resetVFATCRCCounters() REMOVED
T HwOptoHybrid::resetVFATCRCCount() REMOVED
T HwOptoHybrid::resetWBMasterCounters() REMOVED
T HwOptoHybrid::resetWBSlaveCounters() REMOVED
T HwOptoHybrid::updateVFATCRCCounters() REMOVED
T HwOptoHybrid::updateWBSlaveCounters() REMOVED
T HwOptoHybrid::updateWBMasterCounters() REMOVED
T HwOptoHybrid::LinkStatus() REMOVED
T HwOptoHybrid::LinkReset(unsigned char const&) REMOVED
T HwOptoHybrid::stopScanModule(bool, bool) REMOVED
T HwOptoHybrid::getScanStatus(bool) REMOVED
T HwOptoHybrid::sendBXN(unsigned int const&, unsigned int const&) ADDED
T HwOptoHybrid::sendCalPulse(unsigned int const&, unsigned int const&) REMOVED
T HwOptoHybrid::sendL1ACalPulse(unsigned int const&, unsigned int const&, unsigned int const&) REMOVED
W HwOptoHybrid::getT1GeneratorStatus() REMOVED
T HwOptoHybrid::configureT1Generator(unsigned char const&, unsigned char const&, HwOptoHybrid::T1Sequence, bool) REMOVED
T HwOptoHybrid::startT1Generator(unsigned int const&, unsigned int const&, unsigned int const&) REMOVED
T HwOptoHybrid::stopT1Generator(bool) REMOVED
W HwOptoHybrid::resetT1Generator() REMOVED
HwOptoHybrid::setCDCEClock(bool source, bool fallback) REMOVED
HwOptoHybrid::setVFATClock(bool source, bool fallback) REMOVED
HwOptoHybrid::StatusCDCEClock() REMOVED
HwOptoHybrid::StatusVFATClock() REMOVED
HwOptoHybrid::m_disabledMask  REMOVED
HwOptoHybrid::m_connectedMask REMOVED
```

##### `HwGenericAMC` (incomplete)
```
T HwGenericAMC::LinkStatus(unsigned char const&) REMOVED
W HwGenericAMC::ResetLinks(unsigned char const&) REMOVED
T HwGenericAMC::LinkReset(unsigned char const&, unsigned char const&) REMOVED
T HwGenericAMC::getIPBusCounters(unsigned char const&, unsigned char const&) REMOVED
T HwGenericAMC::resetIPBusCounters(unsigned char const&, unsigned char const&) REMOVED
```

## Checklist:


bdorney:
> Possibly XHALInterface should have a singelton as the RPC connection to a given endpoint, such that we don't create an RPC connection per GEMHwDevice that connects to the same hardware

There is a disconnect method in the `wise::RPCSvc` API. So unused connections can be closed.

Additionally from python API experience a single CTP7 can handle between 40-50 open ROC connections before it dies. This is an extreme case and for sure we should never have this tona single card at P5; but it would be beneficial to limit our connection number. But that’s could be done in a different PR

gembase/include/gem/base/GEMApplication.h
```diff
@@ -82,9 +82,9 @@ namespace gem {
         friend class GEMWebApplication;
 
       public:
-        static const unsigned MAX_AMCS_PER_CRATE      = 12;  ///< I SHOULD PROBABLY NOT BE HARD CODED
-        static const unsigned MAX_OPTOHYBRIDS_PER_AMC = 12;  ///< I SHOULD PROBABLY NOT BE HARD CODED
-        static const unsigned MAX_VFATS_PER_GEB       = 24;  ///< I SHOULD PROBABLY NOT BE HARD CODED
+        static constexpr size_t MAX_AMCS_PER_CRATE      = 12;  ///< I SHOULD PROBABLY NOT BE HARD CODED
+        static constexpr size_t MAX_OPTOHYBRIDS_PER_AMC = 12;  ///< I SHOULD PROBABLY NOT BE HARD CODED
+        static constexpr size_t MAX_VFATS_PER_GEB       = 24;  ///< I SHOULD PROBABLY NOT BE HARD CODED
```
lmoureaux, line 87: Would be good candidates to put in a `configure` step before the actual build happens. Also, they should probably be moved to another file (I need them in places that don't depend on `GEMApplication`).
Probably something to keep in mind for another PR.

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -90,123 +83,174 @@ namespace gem {
...
+       * @param command Event type to execute
        */
       void workloopDriver(std::string const& command);
 
-      // work loop call-back functions
       /**
-       * @brief
+       * @brief Workloop task bound to the initialize action
+       * @description On reciept of an initialize command, the workloop driver calls this task
+       *              This task then calls initializeAction
```
lmoureaux, line 95: I can't find `@description` in the Doxygen manual. What does it do? Can it be replaced by a documented command? Found about 20 time in this file.

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -90,123 +83,174 @@ namespace gem {
...
-       * defines the behaviour of the application for each state transition
-       * most will be pure virtual to enforce derived application specific
-       * implementations, common implementations will be merged into the base
-       * application
+      /**
+       * @brief State transitions
+       * @description defines the behaviour of the application for each state transition
+       *              most will be pure virtual to enforce derived application specific
+       *              implementations, common implementations will be merged into the base
+       *              application
```
lmoureaux, line 189: If this block is supposed to document a group of functions, use `@defgroup`, `@{` and `@}`.

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -216,20 +260,21 @@ namespace gem {
       */
 
       /**
-       * @brief stateChanged
+       * @brief Defines behaviour of a change in the FSM
+       * @param fsm reference to the toolbox::fsm::FiniteStateMachine
        *
```
lmoureaux: Empty line.

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -274,35 +320,36 @@ namespace gem {
 
       /**
        * @brief
+       * @returns pointer to the application state infospace tool box
```
lmoureaux: I'd put `Returns a pointer to...` in `@brief` instead so one doesn't need to look at the detailed docs. `@returns` can be left empty since the description is clear.

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -274,35 +320,36 @@ namespace gem {
...
-      bool b_accept_web_commands;  // should we allow state transition commands from the web interface
+      bool b_accept_web_commands;  ///< should we allow state transition commands from the web interface
 
-      toolbox::BSem m_wl_semaphore;      // do we need a semaphore for the workloop?
-      toolbox::BSem m_db_semaphore;      // do we need a semaphore for the database?
-      toolbox::BSem m_cfg_semaphore;     // do we need a semaphore for the config file?
-      toolbox::BSem m_web_semaphore;     // do we need a semaphore for the web access?
-      toolbox::BSem m_infspc_semaphore;  // do we need a semaphore for the infospace?
+      toolbox::BSem m_wl_semaphore;      ///< do we need a semaphore for the workloop?
+      toolbox::BSem m_db_semaphore;      ///< do we need a semaphore for the database?
```
lmoureaux, line 340: It's already handled using a recursive lock from `boost`. I don't know if `toolbox::BSem` is recursive, but it's not needed. Also, the comment doesn't seem to document how the variable is used, and the naming convention is not followed.

gembase/src/common/GEMFSMApplication.cc
```diff
@@ -1128,31 +1128,15 @@ bool gem::base::GEMFSMApplication::reset(toolbox::task::WorkLoop *wl)
 }
 
 /*
-  bool gem::base::GEMFSMApplication::noAction(toolbox::task::WorkLoop *wl)
-  {
+bool gem::base::GEMFSMApplication::noAction(toolbox::task::WorkLoop *wl)
+{
```
lmoureaux, line 1132: Why reformat code if it's commented out?

gembase/src/common/GEMFSMApplication.cc
```diff
@@ -1128,31 +1128,15 @@ bool gem::base::GEMFSMApplication::reset(toolbox::task::WorkLoop *wl)
...
-  bool gem::base::GEMFSMApplication::fail(toolbox::task::WorkLoop *wl)
-  {
-  return false;
-  }
+}
 */
 
-/** transition details*/
 /*
-  void gem::base::GEMFSMApplication::initializeAction()
```
lmoureaux, line 1144: Same

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -6,6 +6,8 @@
 #include <iomanip>
 #include <memory>
 
+#include <boost/algorithm/string.hpp>
```
lmoureaux: I'm surprised that this `#include` is needed in a header file. Build time would be improved if it could be moved to the implementation.

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -503,40 +463,27 @@ namespace gem {
...
       // version.
-      //getters
-      const std::string getControlHubIPAddress()  const { return m_controlHubIPAddress;};
-      const std::string getDeviceIPAddress()      const { return m_deviceIPAddress;    };
-      const std::string getIPBusProtocolVersion() const { return m_ipBusProtocol;      };
-      const std::string getAddressTableFileName() const { return m_addressTable;       };
-
+      /**
+         @brief getters
+      */
```
lmoureaux, line 461: Since it's obvious from their name that these functions are getters, it's probably best to keep it as a non-Doxygen comment.
Also, missing `*` on the second line and wrong alignment for `*/`.

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -503,40 +463,27 @@ namespace gem {
...
-      const uint32_t getControlHubPort() const { return m_controlHubPort;};
-      const uint32_t getIPBusPort()      const { return m_ipBusPort;     };
-
-      //setters, should maybe be private/protected? defeats the purpose?
-      void setControlHubIPAddress( std::string const& ipAddress) { m_controlHubIPAddress = ipAddress; };
-      void setIPBusProtocolVersion(std::string const& version) { m_ipBusProtocol = version; };
-      void setDeviceIPAddress(     std::string const& deviceIPAddr) { m_deviceIPAddress = deviceIPAddr; };
+      /**
+         @brief setters, should maybe be private/protected? defeats the purpose?
+      */
```
lmoureaux, line 468: Not a documentation comment, alignment issues.

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -503,40 +463,27 @@ namespace gem {
...
-
       void setDeviceBaseNode(std::string const& deviceBase) { m_deviceBaseNode = deviceBase; };
-      void setDeviceID(      std::string const& deviceID) { m_deviceID = deviceID; };
-
-      void setControlHubPort(uint32_t const& port) { m_controlHubPort = port; };
-      void setIPBusPort(     uint32_t const& port) { m_ipBusPort = port; };
-
-      /* uhal::HwInterface& getGEMHwInterface() const; */
-
+      void setDeviceID(      std::string const& deviceID)   { m_deviceID = deviceID; };
```
lmoureaux: Extra whitespace.

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -546,15 +493,15 @@ namespace gem {
        * @usage
        *
        */
-      virtual void generalReset();
+      virtual void generalReset()=0;
 
       /**
-       * @brief performs a reset of the GLIB counters
+       * @brief performs a reset of the counters
        *
```
lmoureaux, line 493: Use the opportunity to drop empty lines?

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -563,16 +510,13 @@ namespace gem {
...
        * @param link is the link to perform the reset on
        */
-      virtual void linkReset(uint8_t const& link);
+      virtual void linkReset(uint8_t const& link)=0;
 
-      DeviceErrors m_ipBusErrs;
+      mutable DeviceErrors m_ipBusErrs;
```
lmoureaux, line 508: `mutable` is dangerous. Modifying `m_ipBusErrs` may not change the externally visible behavior of the class (or the behavior of the whole program becomes undefined). But this field is public to begin with...

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -581,10 +525,45 @@ namespace gem {
        * @brief Performs basic setup for the device
        * sets connection details (OBSOLETE)
        * sets logging level to Error
-       * Not inherited
+       * Not inherited, but calls a pure virtual function...
```
lmoureaux: Someone reading the documentation will most likely wonder which virtual function.

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -581,10 +525,45 @@ namespace gem {
...
        **/
       void setup(std::string const& deviceName);
 
+      /**
+       * @brief Performs a check on the RPC response to verify whether there is an `error` key set
+       *        Every method that makes an RPC call *must* use this function to check the response
+       *
+       * @param caller should be the name of the function, only used in the error message
+       **/
+      void checkRPCResponse(std::string const& caller) const {
```
lmoureaux: Should probably not be `inline` and be implemented in the `.cc`.

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -581,10 +525,45 @@ namespace gem {
...
+        if (rsp.get_key_exists("error")) {
```
lmoureaux: Also need to check for `rpcerror`

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -581,10 +525,45 @@ namespace gem {
...
+      }
+
+      /**
+       * @brief Extracts the device parameters from the device name
+       *
+       * @param deviceName should be of the format: gem-shelfXX-amcYY
+       *        optionally including -optohybridZZ
+       * @param index selects which of XX, YY, ZZ to return
+       **/
+      static uint8_t extractDeviceID(std::string const& deviceName, uint8_t const& index) {
```
lmoureaux: * Not a good candidate for an `inline` function (too complex), move implementation to `.cc`.
* Finding the `shelf`, `amc` and `optohybrid` parameters would require parsing the name three times. Why not return `std::vector<uint8_t>`?
* Looks like the implementation also supports an extra `-vfatVV`, why not state it in the docs?

gemhardware/include/gem/hw/GEMHwDevice.h
```diff
@@ -581,10 +525,45 @@ namespace gem {
...
+        std::vector<std::string> subs;
+        boost::split(subs, deviceName, boost::is_any_of("-"));
+        if (index < subs.size()) {
+          return stoull(subs[index].substr(subs[index].find_first_of("0123456789"),2),nullptr,10);
```
lmoureaux: What if the value is larger than 255? What if `find_first_of` can't find a digit?

gemhardware/include/gem/hw/HwGenericAMC.h
```diff
@@ -15,32 +15,17 @@ namespace gem {
...
-         * @var AMCIPBusCounters::TrackingStrobe
-         * TrackingStrobe is a counter for the number of errors on the tracking data link
-         * @var AMCIPBusCounters::TrackingAck
-         * TrackingAck is a counter for the number of errors on the trigger data link
-         * @var AMCIPBusCounters::CounterStrobe
-         * CounterStrobe is a counter for the number of errors on the tracking data link
-         * @var AMCIPBusCounters::CounterAck
-         * CounterAck is a counter for the number of errors on the trigger data link
          */
         typedef struct AMCIPBusCounters {
```
lmoureaux, line 22: Since you're already changing the lines above and below, please drop the extra `typedef`, this is C++, not C.

gemhardware/include/gem/hw/HwGenericAMC.h
```diff
@@ -84,38 +69,48 @@ namespace gem {
...
         /****************************/
         /**
-         * Read the board ID registers
+         * @brief Read the board ID register (should be reimplemented in derived HW, e.g., GLIB)
+         * @param legacy determines whether to read the legacy system board ID (GLIB)
+         *        or the GEM_AMC board ID
          * @returns the AMC board ID as a std::string
          */
-        virtual std::string getBoardID();
+        virtual std::string getBoardID(bool const legacy=false);
```
lmoureaux: Use a pure virtual function to document the need for implementation in derived classes?
I couldn't read the whole PR yet, but I assume there is a hierarchy like:
```
HWGenericAMC
    HWAMC
    HWGLIB
```
I'll let @lpetre-ulb comment, but I don't think there will be any visible difference between the AMC and the GLIB other than the hostname.
Also, `const`?

gemhardware/include/gem/hw/exception/Exception.h
```diff
@@ -115,27 +117,27 @@ GEM_HW_DEFINE_ALARM(MonitoringFailureAlarm)
...
-  msg << MSG_BASE << " caught exception " << e.what();      \
-  CMSGEMOS_ERROR(msg.str());                                \
-  XCEPT_RAISE(EX_TYPE, msg.str());                          \
-} catch (std::exception const& e) {                         \
-  std::stringstream msg;                                    \
-  msg << MSG_BASE << " caught exception " << e.what();      \
-  CMSGEMOS_ERROR(msg.str());                                \
-  XCEPT_RAISE(EX_TYPE, msg.str());                          \
+#define GEM_HW_TRANSITION_CATCH(MSG_BASE, EX_TYPE)             \
+  catch (uhalException const& e) {                             \
```
lmoureaux: Un-indent the first `catch`.

gemhardware/include/gem/hw/glib/GLIBManager.h
```diff
@@ -52,15 +52,13 @@ namespace gem {
           virtual void stopAction()       throw (gem::hw::glib::exception::Exception) override;
           virtual void haltAction()       throw (gem::hw::glib::exception::Exception) override;
           virtual void resetAction()      throw (gem::hw::glib::exception::Exception) override;
-          //virtual void noAction()         throw (gem::hw::glib::exception::Exception) override;
+          // virtual void noAction()         throw (gem::hw::glib::exception::Exception) override;
 
-          virtual void failAction(toolbox::Event::Reference e)
+          void failAction(toolbox::Event::Reference e)
```
lmoureaux, line 57: Is `virtual` removed because this function `override`s? If so add `override`.

gemhardware/include/gem/hw/glib/GLIBManager.h
```diff
@@ -77,36 +75,25 @@ namespace gem {
           void     createGLIBInfoSpaceItems(is_toolbox_ptr is_glib, glib_shared_ptr glib);
           uint16_t m_amcEnableMask;
 
-          toolbox::task::WorkLoop *p_amc_wl; ///< paralelize the calls to different AMCs
-          toolbox::BSem m_amc_wl_semaphore[MAX_AMCS_PER_CRATE]; // do we need a semaphore for the workloop or each of them?
+          toolbox::task::WorkLoop *p_amc_wl;                     ///< paralelize the calls to different AMCs
```
lmoureaux, line 78: * Naming convention
* Why is a raw pointer needed?

gemhardware/include/gem/hw/glib/GLIBManager.h
```diff
@@ -135,20 +113,21 @@ namespace gem {
             };
           };
 
-          mutable gem::utils::Lock m_deviceLock;  // [MAX_AMCS_PER_CRATE];
+          mutable gem::utils::Lock m_deviceLock;  ///< [MAX_AMCS_PER_CRATE];
```
lmoureaux, line 116: This is for sure not a documentation comment.

gemhardware/include/gem/hw/glib/GLIBManager.h
```diff
@@ -21,14 +21,14 @@ namespace gem {
       class GLIBManagerWeb;
       class GLIBMonitor;
 
-      typedef std::shared_ptr<HwGLIB>  glib_shared_ptr;
-      typedef std::shared_ptr<gem::base::utils::GEMInfoSpaceToolBox> is_toolbox_ptr;
+      using glib_shared_ptr = std::shared_ptr<HwGLIB>;
+      using is_toolbox_ptr  = std::shared_ptr<gem::base::utils::GEMInfoSpaceToolBox>;
```
lmoureaux, line 25: Naming convention

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -274,35 +320,36 @@ namespace gem {
...
+       * @returns pointer to the application state infospace tool box
        */
       std::shared_ptr<utils::GEMInfoSpaceToolBox> getAppStateISToolBox() { return p_appStateInfoSpaceToolBox; };
 
     private:
-      GEMFSM m_gemfsm;
+      GEMFSM m_gemfsm;  ///<
 
-      bool b_accept_web_commands;  // should we allow state transition commands from the web interface
+      bool b_accept_web_commands;  ///< should we allow state transition commands from the web interface
```
lmoureaux, line 337: Naming conventions

gembase/include/gem/base/GEMFSMApplication.h
```diff
@@ -274,35 +320,36 @@ namespace gem {
 
       /**
        * @brief
+       * @returns pointer to the application state infospace tool box
        */
       std::shared_ptr<utils::GEMInfoSpaceToolBox> getAppStateISToolBox() { return p_appStateInfoSpaceToolBox; };
 
     private:
-      GEMFSM m_gemfsm;
+      GEMFSM m_gemfsm;  ///<
```
lmoureaux: Naming conventions

lmoureaux (changes requested):
Went through approximately 1/5 of the changes and almost have 30 comments, so stopping. Will add more comments in follow-up reviews.

Overall:
* I appreciate the effort on documentation and code cleanup :100:
* Some issues with Doxygen
* Some issues with the coding conventions :zipper_mouth_face: 
* Some issues in the C++ code :-1:

Since what I've read is mostly cosmetics, I expect more C++ comments to come.

jsturdy:
@lmoureaux fully agree that there is plenty of room for more cleanup, but I would suggest to focus comments on this PR to things that have been added/changed in these commits, rather than things that have been in the codebase for a long time and have only been tweaked or renamed without change of functionality.

This will probably reduce the time spent commenting, though I would definitely still appreciate in your summary a list of the inconsistencies that you noticed, so that they can be addressed in a dedicated "style cleanup" PR

gemhardware/include/gem/hw/HwGenericAMC.h
jsturdy: This hierarchy is historical, but probably before `v2.0.0` all AMC specialization will be removed.
In the past the `GLIB` had a user and system firmware block, and `GEM_AMC` basically was providing the user block, with no promise of continued support of the system block (SRAMs, hardware specific features, etc.), so the `GenericAMC` part corresponds to what `GEM_AMC` provides generically, across all (potentially) supported hardware, and in this specific case, both blocks provided a "board ID"...
I envision having a dedicated discussion on exactly this topic in the semi-near future.

Yes, all getters *should* be `const`, but there are some member variable modifications that *can* happen the way things are currently implemented (re `IPBusErrors` mentioned above) that need to be cleaned up for a more "correct" interface.

gemhardware/include/gem/hw/GEMHwDevice.h
jsturdy: > * Finding the `shelf`, `amc` and `optohybrid` parameters would require parsing the name three times. Why not return `std::vector<uint8_t>`?

It's a possibility... how I use it didn't propose that option.  If the use case exists, probably better to implement your idea of a `GEMHwAddress` `struct`.

> * Looks like the implementation also supports an extra `-vfatVV`, why not state it in the docs?

I am not envisioning supporting a `HwVFAT` type class (which was never really done in the core code anyway), though this is open for *some* discussion.
The reasoning is that minimizing the number of RPC connections (and also `uhal` devices) is in our best interests, and since what we really have are `BackEndDevice` (`AMC`) and `FrontEndDevice` (covering `OptoHybrid`, `GBTx`, `VFAT`, `SCA`).
Current usage in `cmsgemos` managers, tried at one point to have `HwVFAT` devices, but it never really made good sense, since most of what we wanted to do was broadcast type anyway, and now configurations will be streamed per front-end link anyway.

This is something that will have to be addressed when certain python code is migrated, as @bdorney created `HwVFAT` devices for some tasks, but I don't see a problem migrating this.
In the worst case, we can indeed create `HwVFAT` devices, but I would only do it as a return type from a `HwOptoHybrid` function, *only* as a create-use-destroy (resource) model, and probably *only* as an `XHALInterface` (i.e., RPC only, to avoid the address table parsing cost).

gemhardware/include/gem/hw/glib/GLIBManager.h
jsturdy: At some point, it was not trivial to get smart pointers around some `xdaq` types, which is probably why this remained a raw pointer

gembase/include/gem/base/GEMApplication.h
jsturdy: Definitely a different PR.
On the `xdaq` PC, unlike on the AMC, there is almost definitely going to be an expectation to support `GE1/1` and `GE2/1` with the same compiled code, so anything that can, *should* be derived from HW.  This type of functionality is not (yet) fully implemented in the FW, but has been a longstanding request..

An architecture discussion on how best to achieve this, also has relevance for the templated RPC calls, which currently would use the method of compiling in a generation to the library, will likely take place during an upcoming developers' meeting.

gemhardware/include/gem/hw/GEMHwDevice.h
jsturdy: Actually, the limit is 99...

Good question, we'll have to see when someone grossly misuses this...

gemhardware/include/gem/hw/GEMHwDevice.h
lmoureaux: >> Finding the shelf, amc and optohybrid parameters would require parsing the name three times. Why not return std::vector<uint8_t>?
>
> It's a possibility... how I use it didn't propose that option. If the use case exists, probably better to implement your idea of a GEMHwAddress struct.

If Umesh decides to use the naming system in the DB, I'll need this function (and the reverse, ie creating the name) for `gemonlinedb`.

jsturdy: For more background, this function was added to be able to link a `uhal` connection name (which is unique down to the OptoHybrid) with an `RPC` connection name (which is unique *only* to the AMC).
These names are built up from `shelf`/`slot`/`link`, rather than simply creating devices based on a passed in name (which would then be subsequently deconstructed)

> If Umesh decides to use the naming system in the DB, I'll need this function (and the reverse, ie creating the name) for `gemonlinedb`.

We (online software) *should* have the final say viz requirements and usage of these tables in the DB (since we are the consumer).
I fundamentally do **not** want anything in this code base (with the exception of code in `gemonlinedb`) to have *any* dependency one way or another, on the structure or naming convention inside some DB.
Anything that is needed to done to link a DB to the code should be handled by the `gemonlinedb` package alone.

jsturdy:
I've pushed a few new commits that address *some* of the doc changes, as well as some of the `python` stuff.
Still outstanding will be that many functions that are boosted into `python` have default arguments, and need the `boost` wrapper to create the bindings... or a rethink about default arguments in the `c/c++` API.
I'd leave this as a task for someone to pick up, since it will naturally lead towards migrating other python tools to use the new API.
Full disclosure, `amc_info.py` does **not** work out of the box, due to incomplete porting of previous `python`-only functionality (should be part of followup PR):
```
59 [140719675553600] INFO XHALInterface_gem-shelf01-amc02_0 <> - XHAL Logger tuned up
64 [140719675553600] INFO XHALInterface_gem-shelf01-amc02_0 <> - RPC connected
64 [140719675553600] INFO XHALInterface_gem-shelf01-amc02_0 <> - XHAL Interface connected
log4cplus:ERROR No appenders could be found for logger (gem-shelf01-amc02).
log4cplus:ERROR Please initialize the log4cplus system properly.
18 Jun 2019 17:48:04.082 [7ffbda463740] INFO - amc_user_functions::getAMCObject <> - <gempython.core._cmsgemos_gempython.HwGenericAMC object at 0x7ffbc4bb2fc8>: Success!
18 Jun 2019 17:48:04.082 [7ffbda463740] INFO - amc_user_functions::getAMCObject <> - <gempython.core._cmsgemos_gempython.HwGenericAMC object at 0x7ffbc4bb2fc8>: Success!

--=======================================--
  Opening AMC with ID <gempython.core._cmsgemos_gempython.HwGenericAMC object at 0x7ffbc4bb2fc8>
--=======================================--


--=======================================--
-> GEM SYSTEM INFORMATION
--=======================================--

TK_LINK_RX_POLARITY 0x00000000
TK_LINK_TX_POLARITY 0x00000000
BOARD_ID            0x0000beef
BOARD_TYPE          0x00000001
RELEASE             0x00030701
NUM_OF_OH           0x0000000c
USE_TRIG_LINKS      0x00000001

--=======================================--
-> DAQ INFORMATION
--=======================================--

Traceback (most recent call last):
  File "/opt/cmsgemos/bin/amc_info.py", line 88, in <module>
    printSystemTTCInfo(amc)
NameError: name 'printSystemTTCInfo' is not defined
```

gemhardware/include/gem/hw/glib/GLIBManager.h
```diff
@@ -135,20 +113,21 @@ namespace gem {
...
+          std::array<std::shared_ptr<GLIBMonitor>, MAX_AMCS_PER_CRATE> m_glibMonitors;  ///< AMCMonito pointers to be managed
+          std::array<is_toolbox_ptr, MAX_AMCS_PER_CRATE> is_glibs;                      ///< AMC InfoSpace pointers to be managed
 
-          xdata::Vector<xdata::Bag<GLIBInfo> > m_glibInfo;  // [MAX_AMCS_PER_CRATE];
-          xdata::String                        m_amcSlots;
-          xdata::String                        m_connectionFile;
-          xdata::Boolean                       m_uhalPhaseShift; // FIXME OBSOLETE
-          xdata::Boolean                       m_bc0LockPhaseShift;
-          xdata::Boolean                       m_relockPhase;
+          xdata::Vector<xdata::Bag<GLIBInfo> > m_glibInfo;  ///< [MAX_AMCS_PER_CRATE];
```
lmoureaux, line 122: Not a doc comment

lmoureaux:
## About `const`

I think that methods such as `HwOptoHybrid::getVFATMask` should be `const`: provided that the hardware works as intended, they do not alter its visible state. It would be clear that a function taking `const HwOptoHybrid &` would not modify the hardware configuration.

Note that if there are such things as visible remote call counters (or slow control command counters), they should *not* be made available through the "normal" hardware classes or they will break `const`ness. I'm not sure how to best handle such functionality.

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
```diff
@@ -304,532 +99,202 @@ namespace gem {
...
+          uint32_t getFPGAMinVccAux();
 
           /**
-           * Read the VP/VN from the ADC
+           * @brief Read the VP/VN from the ADC
            * @returns uint32_t raw ADC VP/VN value
            */
-          uint32_t getADCVPVN() {
-            return readReg(getDeviceBaseNode(),"ADC.VPVN"); };
+          uint32_t getADCVPVN() const;
```
lmoureaux, line 268: Why is this function `const` while most others are not? If ever implemented it will need to perform the same `readReg()` call. Dropping `const` once the function is implemented might break code. [This is unrelated to my `const` comment.]

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
```diff
@@ -874,325 +339,175 @@ namespace gem {
...
+           * @brief Sends a read request to all (un-masked) VFATs on the same register
            * @param std::string name name of the register to broadcast the request to
            * @param uint32_t mask specifying which VFATs will receive the broadcast command
            * @param bool reset specifying whether to reset the firmware module first
            * @returns a std::vector of uint32_t words, one response for each VFAT
            */
           std::vector<uint32_t> broadcastRead(std::string const& name,
                                               uint32_t    const& mask=gem::hw::utils::ALL_VFATS_BCAST_MASK,
-                                              bool               reset=false);
+                                              bool        const& reset=false);
```
lmoureaux, line 507: Why pass a `bool` by `const` reference? It may actually be slower than a plain `bool` (cleanup: same applies to `mask`).

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
```diff
@@ -1201,82 +516,76 @@ namespace gem {
           void broadcastWrite(std::string const& name,
                               uint32_t    const& value,
                               uint32_t    const& mask=gem::hw::utils::ALL_VFATS_BCAST_MASK,
-                              bool               reset=false);
+                              bool        const& reset=false);
```
lmoureaux, line 519: Same (`bool` by `const` reference)

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
```diff
@@ -1201,82 +516,76 @@ namespace gem {
...
+           * @brief Sends a write request for all configuration registers on all GBTx chips
+           * @param gbtID is the GBTx to which the configuration will be written
+           * @param gbtcfg is a pointer to an array of 92 32-bit words
+           * @param gbtcfg is an array of 92 32-bit words
+           */
 
-          std::vector<linkStatus> getActiveLinks() { return v_activeLinks; }
-          bool isLinkActive(int i) { return b_links[i]; }
+          void configureGBT(uint8_t const& gbtID, uint32_t const* gbtcfg); // FIXME, make private
+          void configureGBT(uint8_t const& gbtID, std::array<const uint32_t, 92> const& gbtcfg);
```
lmoureaux, line 560: * I'm not sure where the 92 comes from, maybe better to use a symbolic constant?
* Why must the contents of the array must always be `const`? (ie why not `std::array<uint32_t, 92> const &`)

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
```diff
@@ -1201,82 +516,76 @@ namespace gem {
...
-          OptoHybridWBSlaveCounters  m_wbSlaveCounters;  /** Wishbone slave transaction counters */
-          OptoHybridT1Counters       m_t1Counters;       /** T1 command counters */
-          OptoHybridVFATCRCCounters  m_vfatCRCCounters;  /** VFAT CRC counters */
+          /**
+           * @brief Sends a write request for all configuration registers on all GBTx chips
+           * @param gbtcfg is a pointer to an array of 3*92 32-bit words
+           * @param gbtcfg is an array of 3*92 32-bit words
+           */
+          void configureAllGBTs(uint32_t const* gbtcfg); // FIXME, make private
+          void configureAllGBTs(std::array<const uint32_t, 3*92> const& gbtcfg);
```
lmoureaux, line 568: It's not clear from the interface how values should be interleaved in the array, ie is it equivalent to
`std::array<const uint32_t[3], 92>` or `std::array<const uint32_t, 92>[3]` or something more fancy. Let me suggest `std::array<gbt_config_data_t, 3>` with `using gbt_config_data_t = std::array<const uint32_t, 92>`.

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
```diff
@@ -1201,82 +516,76 @@ namespace gem {
...
-          uint32_t m_disabledMask;   ///<
-          uint32_t m_connectedMask;  ///<
-          uint8_t m_controlLink;     ///<
-          int m_slot;                ///<
+          bool b_is_initial;         ///< Don't query HW for certain lookups and use cached values FIXME OBSOLETE?
+          uint8_t m_link;            ///< Link on the AMC the OptoHybrid is connected to
+          uint32_t m_disabledMask;   ///< mask where a '1' means ignore data FIXME OBSOLETE?
+          uint32_t m_connectedMask;  ///< mask where a '1' means VFAT is connected FIXME OBSOLETE?
 
+          std::vector<std::pair<uint8_t, uint32_t> > m_chipIDs; ///< mapping of VFAT position to chipID
```
lmoureaux, line 586: Why reorder? This belongs together with `b_is_initial`.

gemhardware/include/gem/hw/optohybrid/OptoHybridManager.h
```diff
@@ -39,208 +39,93 @@ namespace gem {
...
               return os.str();
             };
           };
 
         private:
 
-          void     createOptoHybridInfoSpaceItems(is_toolbox_ptr is_optohybrid, optohybrid_shared_ptr optohybrid);
+          void createOptoHybridInfoSpaceItems(is_toolbox_ptr is_optohybrid, optohybrid_shared_ptr optohybrid);
+
+          mutable gem::utils::Lock m_deviceLock;  ///< [MAX_OPTOHYBRIDS_PER_AMC*MAX_AMCS_PER_CRATE];
```
lmoureaux, line 91: This is not a documentation comment (code leftover).

gemhardware/src/common/glib/GLIBManager.cc
```diff
@@ -206,21 +199,20 @@ void gem::hw::glib::GLIBManager::initializeAction()
...
+    uint32_t slot = info.slotID.value_-1;
     CMSGEMOS_DEBUG("GLIBManager::creating pointer to card in slot " << (slot+1));
 
     // create the cfgInfoSpace object (qualified vs non?)
-    std::string deviceName = info.cardName.toString();
-    if (deviceName.empty())
-      deviceName = toolbox::toString("gem-shelf%02d-amc%02d",
-                                     info.crateID.value_,
-                                     info.slotID.value_);
+    std::string deviceName = deviceName = toolbox::toString("gem-shelf%02d-amc%02d",
```
lmoureaux: Going to crash.

lmoureaux (changes requested):
Went (too quickly) through the rest of the code:

* One must-fix issue (likely crash)
* Some API improvements suggested
* I reduced the amount of style- and documentation-related comments, but there are still many issues, including in new or modified code. It seems that you are sometimes using a 2-columns indent while the standard is 4.
* The large number of cosmetic changes make it difficult to focus on the relevant areas, decreasing the review quality.
* Generally speaking a systematic C++11-ification would help cleaning up the code. Many areas are still C++ 98, some even C.

lmoureaux:
Also Codacy found some new issues and many "possibly new", please check.

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
jsturdy: I was placeholding until a header from `ctp7_modules` is included (which has the constants and types)

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
jsturdy: Either because I felt that the promise should be made that the values would not be changed, or because I was matching a declaration in the `ctp7_modules`, again, this type is expected to be `include`'d from a common `constants` definition

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
jsturdy: I did it for uniformity of intent, with a consideration that a future cleanup may pass by value in the declaration, then in the implementation the signature would be by `const` value, just for the compiler check.
In any event, probably the compiler is smart enough to optimize this...

jsturdy:
> Went (too quickly) through the rest of the code:
> 
> * One must-fix issue (likely crash)

Good catch! Human 1, compiler 0 (well, compiler probably wins in the long run... :-) )

> * Some API improvements suggested

> * I reduced the amount of style- and documentation-related comments, but there are still many issues, including in new or modified code. It seems that you are sometimes using a 2-columns indent while the standard is 4.

I have always (I think) throughout this project used 2 spaces.

> * The large number of cosmetic changes make it difficult to focus on the relevant areas, decreasing the review quality.

Quite :-(

> * Generally speaking a systematic C++11-ification would help cleaning up the code. Many areas are still C++ 98, some even C.

Agreed, I think as part of the move to primarily support `xdaq15` a systematic campaign can be done to address this.  I've been doing incrementals in various changes.

jsturdy:
> Also Codacy found some new issues and many "possibly new", please check.

The "new" issues aren't actually new issues, just happen to be in places where changes were made.

Many of the "potential" issues are related to the `python` code, which will be cleaned up as part of a followup PR.

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
jsturdy: Oversight/desire to have getters `const` (as you rightly surmised above)

gemhardware/src/common/glib/GLIBManager.cc
jsturdy: Yikes!

gemhardware/include/gem/hw/optohybrid/HwOptoHybrid.h
lmoureaux: > In any event, probably the compiler is smart enough to optimize this...

It can't if it can't inline, because the reference is part of the binary interface. To confirm, I dumped the x86_64 assembly for the following code with and without `-DREFERENCE` (`gcc` 7.4.0 at `-O3`):
```C++
#ifdef REFERENCE
[[gnu::noinline]] int f(const bool &x)
#else
[[gnu::noinline]] int f(bool x)
#endif
{
    return x ? 42 : 24;
}
```
With reference:
```asm
_Z1fRKb:
.LFB30:
        .cfi_startproc
        cmp     BYTE PTR [rdi], 0 ; Looks at the LSB of the value pointed to by register RDI
        mov     edx, 42
        mov     eax, 24
        cmovne  eax, edx 
        ret 
        .cfi_endproc
```
Without reference:
```asm
_Z1fb:
.LFB30:
        .cfi_startproc
        cmp     dil, 1 ; Direct access to the LSB of register SI (aka DIL)
        sbb     eax, eax 
        and     eax, -18 
        add     eax, 42
        ret 
        .cfi_endproc
```
While I'm quite sure that the access would be cached in L1 memory, it is still a tiny bit slower.

*[edit: had swapped with and without]*

gembase/src/common/GEMFSMApplication.cc
jsturdy: mental consistency...

gemhardware/include/gem/hw/HwGenericAMC.h
jsturdy: Future cleanup

gembase/src/common/GEMFSMApplication.cc
lmoureaux: Ok if it's for your own mental health I think I'll just let it go...

gemhardware/include/gem/hw/glib/GLIBManager.h
jsturdy: Not inherited (currently commented out in base class)

gemhardware/include/gem/hw/GEMHwDevice.h
jsturdy: Here it was added simply to attempt to apply `const` where generally appropriate. Redesign of the `GEMHwDevice` interface/state probably necessary to do this 100% correctly

gemhardware/include/gem/hw/GEMHwDevice.h
jsturdy: Removed with moving `extractDeviceID` to the implementation

jsturdy:
At this point, I think I've addressed all the docs/cleanup comments I feel are worth addressing in this PR.

After review I will clean up the commits, but I won't be adding any new changes myself.

lmoureaux (dismissed):
Here is a file with C++ code differences only (formatted in LLVM style): [code-differences.txt](https://github.com/cms-gem-daq-project/cmsgemos/files/3307498/code-differences.txt)
It's about half the line count so is a bit easier to read.

Didn't find further significant issues.

PR #296, opened 2019-07-09
Add default GBT RX phases for long GE1/1 GEBs

lpetre-ulb:
## Description

From today long GE1/1 GEBs will be tested at QC7. Therefore a LUT must be available in order to cope with the "all good phases" case during the GBT RX phase scan. Based on OH&GEB QC experience at ULB good default phases can be deduced.

The following plot summarizes the GBT phase scan result for 20 long GEBs. Phases were read by eye from the plot.

LongGEB

### Types of changes

## How Has This Been Tested?

`testConnectivity.py` runs with long GEBs and does not write the place holder phase (0).

## Checklist:


bdorney (approved):

jsturdy:
Is the "by eye" method the same that was used for the other LUTs?  Otherwise, you could feed this metadata into the phase finding algorithm

bdorney:
> Is the "by eye" method the same that was used for the other LUTs? Otherwise, you could feed this metadata into the phase finding algorithm

Yes the original LUT was made by eye as well; but indeed yes we could send the data into the phase finding algorithm.  Do you think we would significantly gain here?

lpetre-ulb:
Moreover there is no straightforward way of using the phase finding algorithm since it is fully embedded inside `testConnectivity.py`. I was thinking to use the algorithm the program requested in https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/232 But in which repository should the algorithm go? Maybe in `gem-plotting-tools` since it is already required by `vfatqc-python-scripts`? How to deal with the "all phases good" then?

mexanick (commented):
cursory

PR #300, opened 2019-07-16
Addressing #299: Changing setDAQParamsPostConfigure.py to Take Sysadmin Configurable link enable mask

### LINKED ISSUE #299: Feature Request: Change setDAQParamsPostConfigure.py to allow Sysadmin Provided Input Enable Mask Per AMC Slot
## Brief summary of issue

Right now `setDAQParamsPostConfigure.py` dynamically determines the `ohMask` that each slot should use when setting the `INPUT_ENABLE_MASK`.  This is convenient but it does prohibit a sysadmin from removing a particular link from data-taking.

We'd like to be able to control which links are used in `INPUT_ENABLE_MASK`.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

I worked out a change that allows a sysadmin to change a set of files:

```bash
$GEM_ADDRESS_TABLE_PATH/gem-shelf01-input-en-mask.txt
$GEM_ADDRESS_TABLE_PATH/gem-shelf02-input-en-mask.txt
...
```

Each file is a comma separated file:

```
#slot,ohMask#
#############
1,0xfff
2,0xfff
3,0xfff
4,0xfff
5,0xfff
6,0xfff
7,0xfff
8,0xfff
9,0xfff
10,0xfff
11,0xfff
12,0xfff
```

And allows a sysadmin to specify the ohMask that should be used for each AMC.  Formats available are hex, bin, or dec.

Then at runtime `setDAQParamsPostConfigure.py` will look for the file matching the appropriate shelf number and then load the ohMask for each AMC slot.  If an AMC slot is not in the text file then and only then will the ohMask be determined dynamically.

Changes can be seen: 

https://github.com/cms-gem-daq-project/cmsgemos/commit/f568ff26059bcb2f69d5fe888e68311ad9d3f710

## Current Behavior

It's not possible to control the `INPUT_ENABLE_MASK` for data taking when calling `setDAQParamsPostConfigure.py` as it's determined dynamically.

## Context

We have 4/2/B detector on QC8 which has some GBT unlock issues; we are keeping this detector OFF for now.  We want to turn it on to use the new found monitoring to understand the issues however GBT1 is known to send garbage that breaks the event building in the CTP7; so if this link is ON and it's SCA is READY and not in CRITICAL_ERROR it will be automatically included in the data-taking when the shifter calls `setDAQParamsPostConfigure.py`.

This is not desired.

bdorney:
## Description
Fixes #299 

### Types of changes

Breaking because `LINK_ENABLE_MASK` will no longer be determined dynamically by default.

## Motivation and Context

See discussion in #299.

## How Has This Been Tested?

On QC8 machine over weekend with runs 196-200; works like a charm.

## Checklist:

### DIFF
--- a/gempython/tools/xdaq/setDAQParamsPostConfigure.py
+++ b/gempython/tools/xdaq/setDAQParamsPostConfigure.py
@@ -7,20 +7,81 @@ def setDAQParams(args):
     from gempython.utils.gracefulKiller import GracefulKiller
     killer = GracefulKiller()
 
+    # Get ENV variable
+    from gempython.utils.wrappers import envCheck
+    envCheck("GEM_ADDRESS_TABLE_PATH")
+    import os
+    gemAddrPath = os.getenv("GEM_ADDRESS_TABLE_PATH")
+
+    # Determine INPUT_ENABLE_MASK
+    dict_ohMask = {}
+    import sys
+    if os.path.isfile("{}/gem-shelf{:02d}-input-en-mask.txt".format(gemAddrPath,args.shelf)):
+        printGreen("Loading existing input enable mask file at: $GEM_ADDRESS_TABLE_PATH/gem-shelf{:02d}-input-en-mask.txt".format(args.shelf))
+        # Load masks
+        inptEnMaskFile = open("{}/gem-shelf{:02d}-input-en-mask.txt".format(gemAddrPath,args.shelf),"r")
+        for row,line in enumerate(inptEnMaskFile):
+            # Skip commented lines
+            if line[0] == "#":
+                continue
+
+            # Parse line
+            line = line.strip("\n")
+            slotList = line.split(",")
+
+            # Get slot
+            try:
+                thisSlot = int(slotList[0],10)
+            except ValueError as err:
+                printRed("Reading $GEM_ADDRESS_TABLE_PATH/gem-shelf{:02d}-input-en-mask.txt: Invalid value given for AMC slot {}.  I was expecting an integer in [1,12] in decimal representation".format(args.shelf,slotList[0]))
+                inptEnMaskFile.close()
+                sys.exit(os.EX_DATAERR)
+
+            # Get ohMask
+            try:
+                dict_ohMask[thisSlot] = int(slotList[1],16)
+            except ValueError as err:
+                printRed("Reading $GEM_ADDRESS_TABLE_PATH/gem-shelf{:02d}-input-en-mask.txt: For AMC slot {} invalid value given for ohMask: {}.  I was expecting an integer [0x0,0xfff] in hexidecimal representation with a leading '0x'".format(args.shelf,thisSlot,slotList[1]))
+                inptEnMaskFile.close()
+                sys.exit(os.EX_DATAERR)
+                pass
+
+            if dict_ohMask[thisSlot] > 0xfff:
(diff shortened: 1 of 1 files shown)


gempython/tools/xdaq/setDAQParamsPostConfigure.py
```diff
@@ -3,24 +3,90 @@
 from gempython.tools.amc_user_functions_uhal import *
 from gempython.utils.gemlogger import printGreen, printRed, printYellow
 
+def convert2Int(inputStr):
```
jsturdy: Generally not a fan of attempts at a generic wrapper to parse a `string` as an `int` because there are plenty of corner cases that should be covered
I prefer that when an argument can be represented in a variety of ways, this should be defined in the usage, and an expectation mentioned there.
Some of the problems here are that:
* `0xbad` is identical to `bad`, but `bad` will be parsed as a binary and will fail.
* Similarly, `0xcafe` is identical to `cafe`, which would be parsed as a decimal and will fail.

gempython/tools/xdaq/setDAQParamsPostConfigure.py
```diff
@@ -3,24 +3,90 @@
...
+            if line[0] == "#":
+                continue
+
+            # Parse line
+            line = line.strip("\n")
+            slotList = line.split(",")
+
+            # Get slot
+            try:
+                thisSlot = convert2Int(slotList[0])
```
jsturdy: Here you should probably make some protection against illegal values.

gempython/tools/xdaq/setDAQParamsPostConfigure.py
```diff
@@ -3,24 +3,90 @@
...
+            except ValueError as err:
+                printRed("Reading $GEM_ADDRESS_TABLE_PATH/gem-shelf{:02d}-input-en-mask.txt: Invalid value given for AMC slot {}".format(args.shelf,slotList[0]))
+                inptEnMaskFile.close()
+                sys.exit(os.EX_DATAERR)
+
+            # Get ohMask
+            try:
+                dict_ohMask[thisSlot] = convert2Int(slotList[1])
```
jsturdy: Likewise here, what prevents an `ohMask` of `0xcafe` of being parsed and applied here?
I guess you probably only use the "valid" bits later, but you could reasonably emit a warning here to at least alert on a parsed value that someone *thought* was OK

jsturdy (changes requested):
Few cautionary comments that could result in changes

bdorney:
Good points made; I’ll try to address these today.

gempython/tools/xdaq/setDAQParamsPostConfigure.py
bdorney: thrown in a check that if this is greater than `0xfff` it's reset to `0xfff`.

gempython/tools/xdaq/setDAQParamsPostConfigure.py
bdorney: removed this function

gempython/tools/xdaq/setDAQParamsPostConfigure.py
bdorney: it's catching a `ValueError` already and I've updated the failure message to state what it was expecting.

bdorney:
Changes addressed and `rebase`d.  Ready for re-review.

jsturdy (approved):

PR #303, opened 2019-08-26
New Feature: Methods for Setting Calibration Data Format in HwAMC & trackingDataCalibrationScan.py

bdorney:
## Description

[GEM_AMC v3.9.0](https://github.com/evka85/GEM_AMC/releases/tag/v3.9.0) introduces a calibration data format to enable a more efficient use of available bandwidth of the DAQ path (e.g. CTP7 -> AMC13 -> FED).

List of changes:
- Provides new method `HwAMC::configureCalDataFormat` for toggling calibration data format.
- Provides FW version protection on using calibration mode in `HwAMC:: configureCalMode`
- `trackingDataCalibrationScan.py` now uses calibration data format by default but has a flag for using the regular GEM data format instead (option `--doNotUseCalDataFormat`).

### Types of changes

Breaking because `trackingDataCalibrationScan.py` now uses calibration data format by default.  However flag `--doNotUseCalDataFormat` exists to use previous (e.g. current) behavior.

## Motivation and Context

We are limited by the 5 Gbps bandwidth between CTP7 -> AMC13 in the uTCA crate and the 10 Gbps bandwidth between AMC13 and uFEDKit (I believe this increases to 20 Gbps for P5?).  This means that we cannot send L1A's at the same rate of the current tools (e.g. ~167kHz) when using the full data path.  The implementation of the calibration data format significantly reduces the amount of data sent per VFAT when in calibration mode making it possible to increase the rate of L1A's.

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1089564

## Checklist:

### DIFF
--- a/gempython/tools/amc_user_functions_xhal.py
+++ b/gempython/tools/amc_user_functions_xhal.py
@@ -176,8 +176,15 @@ class HwAMC(object):
         rpc_connect(self.name)
         self.nOHs = self.readRegister("GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH")
         self.fwVersion = self.readRegister("GEM_AMC.GEM_SYSTEM.RELEASE.MAJOR")
+
+        fwMajor = self.fwVersion
+        fwMinor = self.readRegister("GEM_AMC.GEM_SYSTEM.RELEASE.MINOR")
+        fwBuild = self.readRegister("GEM_AMC.GEM_SYSTEM.RELEASE.BUILD")
+        self.getFullFWVersion = lambda x,y,z : ((x & 0xff) << 16) | ((y & 0xff) << 8) | (z & 0xff)
+        self.fwVersionFull = self.getFullFWVersion(fwMajor, fwMinor, fwBuild)
         if debug:
-            print "My FW release major = ", self.fwVersion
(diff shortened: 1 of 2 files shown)


lpetre-ulb:
Is this PR still relevant? Will the full tracking data path scans with the calibration data format ever be used with the legacy software? @cgalloni, you may be interested to see how it was done. Merging the PR for future reference is probably not a good reason.

mexanick:
I believe this PR is stale. Legacy software is not going to support this and we should reformat this issue for the new project organization

cgalloni:
I think the caldata format was never in use really, but I've be using the  `trackingDataCalibrationScan.py` as example for the calibration suite implementation.

lpetre-ulb:
IIRC, a couple runs have been taken with the calibration data format for further investigation and software compatibility and debugging.

> I believe this PR is stale. Legacy software is not going to support this and we should reformat this issue for the new project organization

I support the closure, except if...

> I think the caldata format was never in use really, but I've be using the `trackingDataCalibrationScan.py` as example for the calibration suite implementation.

you want to use further the `trackingDataCalibratioScan.py` as an example. It is safer to store the code in this repository than in a personal developer account.

That script was never used in production anyway...

PR #314, opened 2020-02-11
Fixing compilation issues related to the xhal package update

### LINKED ISSUE #313: cmsgemos develop doesn't compile against templated version of xhal package
## Brief summary of issue

Due to structural changes of the templated `xhal` package the `gemhwdevices` and `gemdaqmonitor` packages can't compile. Changes are required in headers and makefiles mostly and are quite trivial. 
### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Possible Solution (for bugs)

Fix will be provided shortly in a corresponding pull request
## Your Environment

* Version used:
develop branch of `cmsgemos`
```bash
xhal
Arch        : x86_64
Version     : 4.0.0
Release     : 0.1.1.374.alpha1.15f6afbgit.centos7.gcc8_3_1
```
* Shell used:
bash

mexanick:
## Description

Fixes #313 
### Types of changes

## How Has This Been Tested?

Compiles on xdaq15 VM
## Checklist:

### DIFF
--- a/gemhardware/devices/Makefile
+++ b/gemhardware/devices/Makefile
@@ -57,7 +57,7 @@ LibraryDirs+=$(BUILD_HOME)/gemutils/lib/$(XDAQ_OS)/$(XDAQ_PLATFORM)
 LibraryDirs+=$(uHALROOT)/lib
 LibraryDirs+=$(XHAL_ROOT)/lib
 
-DependentLibraries =cactus_uhal_uhal xhal
+DependentLibraries =cactus_uhal_uhal xhal-client
 DependentLibraries+=gemutils gemhwutils
 
 include $(XDAQ_ROOT)/$(BUILD_SUPPORT)/Makefile.rules
--- a/gemdaqmonitor/Makefile
+++ b/gemdaqmonitor/Makefile
@@ -47,7 +47,7 @@ DependentLibraryDirs+=$(XHAL_ROOT)/lib
 UserCFlags+=$(PYTHONCFLAGS)
 
 DependentLibraries =gembase
-DependentLibraries+=gemhwdevices gemhwutils xhal
+DependentLibraries+=gemhwdevices gemhwutils xhal-client
 DependentLibraries+=boost_iostreams
 
 include $(XDAQ_ROOT)/$(BUILD_SUPPORT)/Makefile.rules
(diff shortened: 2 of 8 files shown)


lpetre-ulb (approved):
Cursory approval.

gemhardware/devices/Makefile
```diff
@@ -57,7 +57,7 @@ LibraryDirs+=$(BUILD_HOME)/gemutils/lib/$(XDAQ_OS)/$(XDAQ_PLATFORM)
 LibraryDirs+=$(uHALROOT)/lib
 LibraryDirs+=$(XHAL_ROOT)/lib
 
-DependentLibraries =cactus_uhal_uhal xhal
+DependentLibraries =cactus_uhal_uhal xhal-client
```
jsturdy, line 60: does it not also depend on the `xhal-base` library?

lpetre-ulb: What do you mean by "it"? The PR or the `xhal-client` library? If you are referring to library, the implicit dependency shouldn't be added here (this is not a static library). If you are referring to the PR, yes. However the `common` namespace is misleading with respect to the library `-base` suffix.

jsturdy: > What do you mean by "it"? The PR or the `xhal-client` library? If you are referring to library, the implicit dependency shouldn't be added here (this is not a static library). If you are referring to the PR, yes. However the `common` namespace is misleading with respect to the library `-base` suffix.

"it" == `libgemhardware.so`

also, the rpm package is simply `xhal` :-D (though it was mentioned that this could be a dummy package and the actual `xhal-base` library and friends could be packaged as `xhal-base`...

mexanick: actually I don't think it should depend on base - why? It doesn't actually use `base` functionality, it uses the `client` one. Moreover, `base` only goes to `arm` in *glibless* world

jsturdy: > actually I don't think it should depend on base - why? It doesn't actually use `base` functionality, it uses the `client` one. Moreover, `base` only goes to `arm` in _glibless_ world

`base` provides the core RPC translator functionality, so it winds up on both ends of the connection

lpetre-ulb: _If_ you don't explicitly use `base` classes or functions in the code, you shouldn't have to depend on it (or else the build system should be considered as broken).

In this PR, classes inside `xhal::common::utils` _are_ explicitly used; nothing guarantees that `xhal-client` will _always_ depends on `xhal` (without `-base` :-D)

PR #318, opened 2020-02-28
[release prep] v1.0.1

jsturdy:
* Bump metadata for 1.0.1
* Fix python packaging consistency syntax

Once merged, merged commit should be tagged `v1.0.1`

### DIFF
--- a/gempython/Makefile
+++ b/gempython/Makefile
@@ -20,7 +20,7 @@ $(info PythonModules=${PythonModules})
 
 GEMPYTHON_VER_MAJOR=1
 GEMPYTHON_VER_MINOR=0
-GEMPYTHON_VER_PATCH=0
+GEMPYTHON_VER_PATCH=1
 
 include $(BUILD_HOME)/$(Project)/config/mfDefsGEM.mk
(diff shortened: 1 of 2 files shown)


lpetre-ulb (approved):

lpetre-ulb:
Trivial change.

Related question, but probably not so important for the legacy releases: how are the tags supposed to work if all xDAQ libraries have their own version?

jsturdy:
> Related question, but probably not so important for the legacy releases: how are the tags supposed to work if all xDAQ libraries have their own version?

Nominally, the `git` tags would track a `cmsgemos` "release" (at some point in the past I considered having every subdirectory be a submodule with its own set of tags, which is more correct...)
It's also mostly an inherited artifact of how xdaq was organized when they used svn.
Could eventually decide that all packages get the same tag version with each build and not bother with the xdaq versioning (but would have to validate that things all worked as they should)
Then the only thing is that for many releases, most of the binaries aren't going to change (which is another potential argument against the monorepo philosophy, depending on implementation)

anyway, more than you asked for :-D

PR #327, opened 2020-03-19
Hw access optimization

### LINKED ISSUE #316: FSM actions refactoring
## Brief summary of issue

GEM's are taking way too long to initialize, while configuration is done very promptly (in seconds).
Possible reasons are:
- abundance of hardware accesses during `initializeAction(...)` in various hardware managers
- issues with uHAL address tables 
  - may consider to remove uHAL dependencies for AMC and OH altogether
### Types of issue

- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)
- [x] Refactoring request

## Expected Behavior

Should take much less time at the `initializeAction` and in general init+conf step has to take less time

## Possible Solution (for bugs)

Suggestions: 
- Do not touch real hardware during the `initializeAction(...)` apart from creating `rpc` connections. All HW accesses to registers should be postponed till configuration step. 
- Since the single-register access are rare, consider the possibility to avoid invocation of uHAL HW access and move completely to RPC-based functionality
- Further optimizations (out of scope of this issue): provide resource sharing for the RPC connections.
  - At least, share LMDB among different connections
  - Eventually use shared connections to optohybrids (one per AMC)

This issue header will be updated with code snippets proposals. Please post your comments/suggestions below, each subthread will be summarized here as well.

mexanick:
## Description

This PR brings:
- Complete removal of uhal dependencies everywhere in `gemhardware` except `amc13`-related code
- Conversion of all the `RPC` calls in `gemhardware` to templated ones
- Partial removal of obsolete functionality where it is obvious
  - hardware monitoring classes are removed as not used 
- Cleaning of the FM transitions

The quality of the code is not up to industrial standards, but this is one more technical stop on the way to a better structured clean project. A number of issues were spotted during the work on this PR which have to be fixed before we can advance. 

Notably:
  - namespaces accross the  and  has to be uniformized
  - Compile-time definition for `GEM_VARIANT` added to makefiles of `gemhardware/managers` and `gemhardware/devices`
    - must be fixed with new build system. Eventually has to be converted into a runtime constant as we want the same backend software to work with different generation front-ends
  - lots of obsolete functions which require interface redesign
  - argument passing between  and  has to be uniformized - see #173 of `ctp7_modules`

Cleaning the FSM transition, getting rid of `uhal` library.
### Types of changes

## How Has This Been Tested?

Not yet tested on the hardware. Help is needed on building/deployment instructions for `ctp7_modules`. So far build system is unclear. However the code compiles and in general should run. Would also prefer a code review before stepping into testing phase.
## Checklist:


jsturdy (commented):
Just a few quick comments before it's ready for reveiw:
* would be better to replace all the `// Deprecated` comments with `[[deprecated]]` or `__attribute__((__deprecated__))` (if the functions themselves aren't going to be removed in this rework)
* why are we reintroducing the duplication of remote vs local function calls? if the method in the CTP7 modules is desired outside of the modules themselves, then why not just export it directly, i.e., why duplicate `readReg` with the `readRemoteReg` wrapper, when the templating was introduced with one of the intended goals of removing this duplication?

I think everything else can wait for proper review

lpetre-ulb:
>     * would be better to replace all the `// Deprecated` comments with `[[deprecated]]` or `__attribute__((__deprecated__))` (if the functions themselves aren't going to be removed in this rework)

:+1: I would even remove the deprecated placeholder functions since they can lead to unexpected misbehavior (if `cmsgemos` still compiles without them).

>     * why are we reintroducing the duplication of remote vs local function calls? if the method in the CTP7 modules is desired outside of the modules themselves, then why not just export it directly, i.e., why duplicate `readReg` with the `readRemoteReg` wrapper, when the templating was introduced with one of the intended goals of removing this duplication?

Wouldn't these remote RPC methods be only temporary until we no longer need raw register accesses from `cmsgemos`? And replacing all function calls by functor calls in the `ctp7_modules` is not going to be pretty...

jsturdy:
> > * why are we reintroducing the duplication of remote vs local function calls? if the method in the CTP7 modules is desired outside of the modules themselves, then why not just export it directly, i.e., why duplicate `readReg` with the `readRemoteReg` wrapper, when the templating was introduced with one of the intended goals of removing this duplication?
> 
> Wouldn't these remote RPC methods be only temporary until we no longer need raw register accesses from `cmsgemos`? And replacing all function calls by functor calls in the `ctp7_modules` is not going to be pretty...

I don't know, either there's a use case for having these low level functions exposed on on the PC side, in which case I'd just export them as is and do a refactor in `ctp7_modules` now (it's a pretty simple regex replace to get it all done), *or* we don't want them outside, in which case why not just make **that** change now, i.e., create the `ctp7_modules` functions that replace the ones that are using these one-off calls. That's what I did with the `HwGenericAMC` -> `amc` module (although I didn't replace the calls in `cmsgemos` at the time).
The third alternative is to have the exported name the same as the "local" name, as they'll be in a different namespace, right?

lpetre-ulb:
> I don't know, either there's a use case for having these low level functions exposed on on the PC side, in which case I'd just export them as is and do a refactor in `ctp7_modules` now (it's a pretty simple regex replace to get it all done), _or_ we don't want them outside, in which case why not just make **that** change now, i.e., create the `ctp7_modules` functions that replace the ones that are using these one-off calls. That's what I did with the `HwGenericAMC` -> `amc` module (although I didn't replace the calls in `cmsgemos` at the time).

I think the use case is only a temporary compatibility layer for migration purposes. I prefer to see multiple easily reviewable small PR rather than one huge PR with many changes. The mid-term goal could be like:

1. Get rid of uHAL (this PR)
2. Create RPC methods that replace the direct read/write accesses (one or more PR)
3. Create a RPC module for monitoring purposes (one or more PR)

> The third alternative is to have the exported name the same as the "local" name, as they'll be in a different namespace, right?

For now, at least, they are in the same namespace, in this case, `utils`. But, the function would hide the class, so the behavior would be the one you expect, i.e. the function available for the `ctp7_modules` and the functor available for remote calls. That's terrible design though.

mexanick:
please check updated PR description

gemhardware/managers/src/common/amc/AMCManager.cc
```diff
@@ -129,33 +129,11 @@ void gem::hw::amc::AMCManager::init()
...
-      // set the web view to be empty or grey
-      // if (!info.present.value_) continue;
-      // needs .value_?
-      // p_gemWebInterface->amcInSlot(slot);
-    }
-  }
-
-  // FIXME make me more streamlined
+  // this maybe shouldn't be done? Commenting out for now, but needs testing
+  //info.slotID  = slot+1;
```
lpetre-ulb, line 132: To be tested. A changelog, or announcement, is required if the configuration has to be changed somehow.

gemhardware/managers/src/common/amc/AMCManager.cc
```diff
@@ -188,74 +166,14 @@ void gem::hw::amc::AMCManager::initializeAction()
       CMSGEMOS_DEBUG("AMCManager::obtaining pointer to HwGLIB");
       m_amcs.at(slot) = amc_shared_ptr(new gem::hw::glib::HwGLIB(deviceName, m_connectionFile.toString()));
       amc_shared_ptr amc = m_amcs.at(slot);
-      if (amc->isHwConnected()) {
-        CMSGEMOS_DEBUG("AMCManager::Creating InfoSpace items for AMC device " << deviceName);
-
-        // FIXME should not need this here?
-        amc->disableDAQLink();
```
lpetre-ulb, line 195: To be tested, since it changes the behavior.

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
```diff
@@ -9,13 +9,11 @@
...
 #include "toolbox/string.h"
 
-#include "uhal/uhal.hpp"
-/* #include "uhal/Utilities.hpp"  /\* removed in 2.5 or 2.6 *\/ */
-#include "uhal/utilities/bits.hpp"
-#include "uhal/utilities/files.hpp"
-
 #include "xhal/client/XHALInterface.h"
-/* #include "xhal/XHALDevice.h" */
+#include "xhal/common/rpc/call.h"
```
lpetre-ulb, line 13: Not technically part of this PR, but wouldn't it be better to have the templatized `call` function as a method of the `XHALInterface`?

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
```diff
@@ -269,209 +81,19 @@ namespace gem {
...
-      void zeroFIFO(std::string const& regName);
-
-
       // These methods provide access to the member variables
       // specifying the `uhal` address table name and the IPbus protocol
       // version.
       /**
          FIXME getters
       */
       const std::string getAddressTableFileName() const { return m_addressTable;   };
```
lpetre-ulb: Is the `m_addessTable` variable still used? Isn't it a IPBus leftover?

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
```diff
@@ -554,30 +170,23 @@ namespace gem {
...
+       * @details Provides basic read/write functionality, reimplemented in daughter classes
        *
        * @usage FILLME
        *
        * @param should reconnect in the case of an already connected manager
        */
-      virtual void connectRPC(bool reconnect=false)=0;
+      virtual void connectRPC(bool reconnect=false);
 
       std::string m_addressTable;   ///< FILLME
```
lpetre-ulb: Leftover?

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
```diff
@@ -269,209 +81,19 @@ namespace gem {
...
          FIXME getters
       */
       const std::string getAddressTableFileName() const { return m_addressTable;   };
-      const std::string getDeviceBaseNode()       const { return m_deviceBaseNode; };
       const std::string getDeviceID()             const { return m_deviceID;       };
 
       /**
          FIXME setters, should maybe be private/protected? defeats the purpose?
       */
       void setAddressTableFileName(std::string const& name) { m_addressTable = "file://${GEM_ADDRESS_TABLE_PATH}/"+name; };
```
lpetre-ulb: Leftover?

gemhardware/devices/src/common/GEMHwDevice.cc
```diff
@@ -57,770 +24,56 @@ gem::hw::GEMHwDevice::~GEMHwDevice()
...
-      errmsg << "Could not read register '" << name << "' (std): "
-             << err.what();
-      CMSGEMOS_ERROR("GEMHwDevice::" << errmsg.str());
-      // XCEPT_RAISE(gem::hw::devices::exception::HardwareProblem, errmsg.str());
-    }
+  uint32_t res = 0;
+  try {
+    res = xhal::common::rpc::call<::utils::readRemoteReg>(rpc, name);
+  } catch (std::exception const& err) {
+    //FIXME
```
lpetre-ulb: Can any read/write error be at least logged?

gemhardware/devices/src/common/GEMHwDevice.cc
```diff
@@ -57,770 +24,56 @@ gem::hw::GEMHwDevice::~GEMHwDevice()
...
-    } catch (std::exception const& err) {
-      std::string msgBase = toolbox::toString("Could not read block '%s' (std)", name.c_str());
-      std::string msg     = toolbox::toString("%s: %s.", msgBase.c_str(), err.what());
-      CMSGEMOS_ERROR("GEMHwDevice::" << msg);
-      // XCEPT_RAISE(gem::hw::devices::exception::HardwareProblem, msg);
-    }
+  try {
+    xhal::common::rpc::call<::utils::writeRemoteReg>(rpc, name, val);
+  } catch (std::exception const& err) {
+    //FIXME
```
lpetre-ulb: Same here.

gemhardware/devices/src/common/amc/HwGenericAMC.cc
```diff
@@ -2,85 +2,33 @@
...
-gem::hw::HwGenericAMC::~HwGenericAMC()
-{
-}
-
-void gem::hw::HwGenericAMC::connectRPC(bool reconnect)
+void gem::hw::amc::HwGenericAMC::connectRPC(bool reconnect)
 {
   if (isConnected) {
     // TODO: find better way than hardcoded versions
+    this->loadModule("utils", "utils v1.0.1"); //FIXME required as atomic transactions are still used here
```
lpetre-ulb: Is there an explicit use of this RPC module in `HwGenericAMC`? If no, I would suggest to simply call `connectRPC` from the parent class.

gemhardware/devices/src/common/optohybrid/HwOptoHybrid.cc
```diff
@@ -81,25 +25,10 @@ gem::hw::optohybrid::HwOptoHybrid::~HwOptoHybrid()
...
-//void gem::hw::optohybrid::HwOptoHybrid::configureDevice()
-//{
-//  // determine the manner in which to configure the device (XML or DB parameters)
-//}
-//
-
 void gem::hw::optohybrid::HwOptoHybrid::connectRPC(bool reconnect)
 {
   if (isConnected) {
+    this->loadModule("utils", "utils v1.0.1");
```
lpetre-ulb: Same here.

gemhardware/devices/src/common/optohybrid/HwOptoHybrid.cc
```diff
@@ -325,43 +213,27 @@ void gem::hw::optohybrid::HwOptoHybrid::setVFATsToDefaults(std::map<std::string,
...
+  //    rsp = rpc.call_method(req);
+  //  } STANDARD_CATCH;
+  //  checkRPCResponse("HwOptoHybrid::configureVFATs");
+  //} GEM_CATCH_RPC_ERROR("HwOptoHybrid::configureVFATs", gem::hw::devices::exception::Exception);
 }
 
 
 void gem::hw::optohybrid::HwOptoHybrid::configureGBT(uint8_t const& gbtID, uint32_t const* gbtcfg)
 {
+  gbt::config_t t_gbtconfig = {*(uint8_t*) gbtcfg};
```
lpetre-ulb: I don't like this kind of "casting"... the `gbtcfg` parameter should directly be of type `gbt::config_t`.

gemhardware/devices/src/common/optohybrid/HwOptoHybrid.cc
```diff
@@ -576,43 +450,15 @@ void gem::hw::optohybrid::HwOptoHybrid::configureScanModule(uint8_t const& mode,
...
-    try {
-      rsp = rpc.call_method(req);
-    } STANDARD_CATCH;
-    checkRPCResponse("HwOptoHybrid::broadcastWrite");
+    xhal::common::rpc::call<::oh::configureScanModule>(rpc, static_cast<uint32_t>(m_link), chip, mode, useUltra, chip, channel, nevts, min, max, step); //FIXME chip is used twice, need a redesign. Probably won't work, but will compile at the moment.
   } GEM_CATCH_RPC_ERROR("HwOptoHybrid::startScanModule", gem::hw::devices::exception::Exception);
 }
 
+//FIXME review passed arguments: nevts is not used here.
 void gem::hw::optohybrid::HwOptoHybrid::startScanModule(uint32_t const& nevts, bool useUltra)
```
lpetre-ulb: Everything related to v2 electronics scans should be removed, not ported to the templated RPC framework. Since it is done, guess it can be postponed to another PR.

gemhardware/managers/src/common/optohybrid/OptoHybridManager.cc
```diff
@@ -262,95 +177,11 @@ void gem::hw::optohybrid::OptoHybridManager::configureAction()
...
-        // FIXME, should not be here or done like this
-        uint32_t gtxMask = optohybrid->readReg("GEM_AMC.DAQ.CONTROL.INPUT_ENABLE_MASK");
-        std::stringstream msg;
-        msg << "OptoHybridManager::configureAction::OptoHybrid connected on link " << static_cast<uint32_t>(link)
-            << " to AMC in slot " << static_cast<uint32_t>(slot+1) << " found, INPUT_ENABLE_MASK changed from "
-            << std::hex << gtxMask << std::dec;
-        gtxMask   |= (0x1<<link);
-        inputMask |= (0x1<<link);
-        // optohybrid->writeReg("GEM_AMC.DAQ.CONTROL.INPUT_ENABLE_MASK", gtxMask);
-        optohybrid->writeReg("GEM_AMC.DAQ.CONTROL.INPUT_ENABLE_MASK", inputMask);
```
lpetre-ulb, line 351: Where is it done now?

lpetre-ulb (commented):
Nothing outstanding, can be compiled. Should be good for hardware tests. I would be careful where and how the `INPUT_ENABLE_MASK` is set.

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
mexanick: Yes, this would be more natural in my opinion and will simplify the includes

gemhardware/managers/src/common/optohybrid/OptoHybridManager.cc
mexanick: nowhere. This should not be done as it was implemented before and should come with a revised configuration methods on the CTP7 side. And certainly should be called on per-OH basis, this is an AMC register, should be set through AMC manager.

lpetre-ulb: This is clearly part of the AMC address space and should be treated as such. But, once your PR will be merged, what happens if someone tries to start a run? Will any chamber be readout? How is it set? (I'm not sure I was around at the time "as it was implemented before".)

mexanick: no, the chambers won't be readout, unless the register "GEM_AMC.DAQ.CONTROL.INPUT_ENABLE_MASK" is correctly written (binary mask which enables corresponding chambers). "as before" means each connected OH adds its "1" to the mask and register is written N(OH) times. This register can be set manually at the moment, but I can foresee quite trivial update to daq configuration module of the CTP7 which will take care of it. However this is beyond the scope of the current PR. If you want this functionality in its obsolete form right now, I can add it back.

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
lpetre-ulb: Issue opened here: https://github.com/cms-gem-daq-project/xhal/issues/149

gemhardware/managers/src/common/optohybrid/OptoHybridManager.cc
lpetre-ulb: Okay, thank you for the precision. That's fine to me for now. @cgalloni, you might be interested by the behavioral change.

gemhardware/devices/src/common/optohybrid/HwOptoHybrid.cc
mexanick: Since the method is not actually used, I changed the API signature to have the `gbt::config_t` as native parameter here

gemhardware/managers/src/common/amc/AMCManager.cc
mexanick: Indeed, but the main purpose if this PR is to have something which uses the templated RPC framework and can cycle through the FSM states. Proper implementation of the state transitions IMHO deserves another PR (and at the moment it has to be also chained with the corresponding changes of the `ctp7_modules`). As a lot of things are candidates for refactoring, simply making it work "as is but without uHAL" doesn't appeal too much to me

gemhardware/devices/src/common/optohybrid/HwOptoHybrid.cc
```diff
@@ -228,19 +228,13 @@ void gem::hw::optohybrid::HwOptoHybrid::configureVFATs(std::map<std::string, uin
...
 }
 
-
-void gem::hw::optohybrid::HwOptoHybrid::configureGBT(uint8_t const& gbtID, uint32_t const* gbtcfg)
+void gem::hw::optohybrid::HwOptoHybrid::configureGBT(uint8_t const& gbtID, std::array<const uint8_t, ::gbt::CONFIG_SIZE> const& gbtcfg)
 {
-  gbt::config_t t_gbtconfig = {*(uint8_t*) gbtcfg};
+  configureGBT(gbtID, gbtcfg.data());
   try {
     xhal::common::rpc::call<::gbt::writeGBTConfig>(rpc, static_cast<uint32_t>(m_link), static_cast<uint32_t>(gbtID), t_gbtconfig);
```
lpetre-ulb: Does not compile: `configureGBT(gbtID, gbtcfg.data());` is not type compliant, would create an inifinte loop and is useless. `t_gbtconfig`, on line 235, is no longer a variable, replaced by `gbtcfg`.

gemhardware/managers/src/common/amc/AMCManager.cc
lpetre-ulb: In general, I believe that regressions should not be allowed in PR. However, this PR as well as the next round of PR are kind of special due to the huge refactoring in progress. But what about the other developments which happen in parallel? Once this PR gets merged, how would the calibration suite be tested, for example?

gemhardware/devices/src/common/optohybrid/HwOptoHybrid.cc
mexanick: ooups, missed the change in the commit. My bad, sorry. Fixed now

mexanick:
**Test report:**

- templated RPC calls could be successfully executed from the cmsgemos
- linking problem against `xhal-base` has been discovered, which prevented correct exceptions treatment and subsequent code crash. Issue reported in [xhal](https://github.com/cms-gem-daq-project/xhal/issues/150) repository, suggested there temporary fix is implemented.
- discovered quite a few attempts to read registers which are not leaf nodes and do not have "read" permissions. This part of the code wasn't changed in the current PR and I suspect that the problem was hidden by `ipbus` transactions ignoring register permissions. Since these bugs don't affect state transitions, I propose to address them in subsequent PR
- FSM transition between "Initial" and "Halted" states is executed correctly. If the HW is present, the system goes to "Halted" state, if HW is not accessible, the system goes to "Error"
- FSM transition between "Halted" and "Configured" state resulted in "Error" due to HW problem
- "Reset" action from "Halted" state hangs up indefinitely

Following this I'd propose to merge this PR and address remaining and future issues in separate PRs. As a proof of concept, the system can work without `uhal` and with templated `rpc`.

gemhardware/devices/Makefile
```diff
@@ -55,7 +59,8 @@ LibraryDirs+=$(BUILD_HOME)/gemutils/lib/$(XDAQ_OS)/$(XDAQ_PLATFORM)
 LibraryDirs+=$(uHALROOT)/lib
```
jsturdy: is this still needed?

gemhardware/devices/Makefile
```diff
@@ -55,7 +59,8 @@ LibraryDirs+=$(BUILD_HOME)/gemutils/lib/$(XDAQ_OS)/$(XDAQ_PLATFORM)
...
 LibraryDirs+=$(XHAL_ROOT)/lib
 
-DependentLibraries =cactus_uhal_uhal xhal-client
+#FIXME hacky linking of xhal-base
+DependentLibraries =cactus_uhal_uhal xhal-client xhal-base
```
jsturdy: and if not, can't you remove `cactus_uhal_uhal`

gemhardware/devices/Makefile
```diff
@@ -44,6 +47,7 @@ IncludeDirs+=$(BUILD_HOME)/gemhardware/utils/include
 IncludeDirs+=$(BUILD_HOME)/gemutils/include
 IncludeDirs+=$(uHALROOT)/include
```
jsturdy: Likewise, still refs `uhal`

gemhardware/devices/include/gem/hw/devices/GEMHwDevice.h
jsturdy: isn't this a reason that `xhal-base` needs to be linked here?

gemhardware/devices/include/gem/hw/devices/amc/HwGenericAMC.h
```diff
@@ -4,774 +4,702 @@
 #define GEM_HW_HWGENERICAMC_H
 
 #include "gem/hw/devices/GEMHwDevice.h"
+#include "ctp7_modules/common/amc.h"
+#include "ctp7_modules/common/amc/daq.h"
+#include "ctp7_modules/common/amc/ttc.h"
```
jsturdy: are these includes needed here? or can they be moved to the implementation?

gemhardware/devices/include/gem/hw/devices/amc/HwGenericAMC.h
```diff
@@ -4,774 +4,702 @@
...
+          virtual bool isHwConnected() override;
+
+          /****************************/
+          /** GEM system information **/
+          /****************************/
+          /**
+           * @brief Read the board ID register (should be reimplemented in derived HW, e.g., GLIB)
+           * @returns the AMC board ID as a std::string
+           */
+          virtual std::string getBoardIDString();
```
jsturdy: is anything inheriting from this?
if not, should remove all `virtual` keywords, and mark `override` those methods which are inherited from `GEMHwDevice`

gemhardware/devices/src/common/amc/HwGenericAMC.cc
```diff
@@ -293,312 +218,249 @@ bool gem::hw::HwGenericAMC::linkCheck(uint8_t const& gtx, std::string const& opM
 }
 
 // FIXME UPDATE //
-// gem::hw::GEMHwDevice::OpticalLinkStatus gem::hw::HwGenericAMC::LinkStatus(uint8_t const& gtx)
+// gem::hw::GEMHwDevice::OpticalLinkStatus gem::hw::amc::HwGenericAMC::LinkStatus(uint8_t const& gtx)
```
jsturdy, line 224: as long as everything's being updated, why not just remove this?
it's clearly not necessary for compatibility with any functionality

gemhardware/devices/src/common/amc/HwGenericAMC.cc
```diff
@@ -293,312 +218,249 @@ bool gem::hw::HwGenericAMC::linkCheck(uint8_t const& gtx, std::string const& opM
...
-        }
-      } STANDARD_CATCH;
-    } STANDARD_CATCH;
-  } GEM_CATCH_RPC_ERROR("HwGenericAMC::enableDAQLink", gem::hw::devices::exception::Exception);
+    xhal::common::rpc::call<::amc::daq::disableDAQLink>(rpc);
+  } GEM_CATCH_RPC_ERROR("HwGenericAMC::disableDAQLink", gem::hw::devices::exception::Exception);
 }
 
-void gem::hw::HwGenericAMC::setZS(bool en)
+void gem::hw::amc::HwGenericAMC::setZS(bool en)
```
jsturdy, line 292: don't all these functions following have `ctp7_modules` equivalents?
why only half update the codebase?
or is this part of the 
> * lots of obsolete functions which require interface redesign

gemhardware/devices/src/common/amc/HwGenericAMC.cc
```diff
@@ -608,343 +470,257 @@ void gem::hw::HwGenericAMC::setDAQLinkRunParameter(uint8_t const& parameter, uin
...
   if (single)
-    return readReg(getDeviceBaseNode(), "TTC.STATUS.TTC_SINGLE_ERROR_CNT");
+    return readReg("GEM_AMC.TTC.STATUS.TTC_SINGLE_ERROR_CNT");
   else
-    return readReg(getDeviceBaseNode(), "TTC.STATUS.TTC_DOUBLE_ERROR_CNT");
+    return readReg("GEM_AMC.TTC.STATUS.TTC_DOUBLE_ERROR_CNT");
 }
 
-uint32_t gem::hw::HwGenericAMC::getTTCCounter(AMCTTCCommandT const& cmd)
+uint32_t gem::hw::amc::HwGenericAMC::getTTCCounter(AMCTTCCommandT const& cmd)
```
jsturdy, line 581: but in cases like this, where there is much more than just a single `(read|write)Reg` it would have made sense to use the `ctp7_modules` functor call

gemhardware/devices/src/common/amc/HwGenericAMC.cc
```diff
@@ -608,343 +470,257 @@ void gem::hw::HwGenericAMC::setDAQLinkRunParameter(uint8_t const& parameter, uin
...
 
-uint32_t gem::hw::HwGenericAMC::getOptoHybridDebugLastCluster(uint8_t const& oh, uint8_t const& cs)
+uint32_t gem::hw::amc::HwGenericAMC::getOptoHybridDebugLastCluster(uint8_t const& oh, uint8_t const& cs)
 {
-  return readReg(getDeviceBaseNode(), toolbox::toString("TRIGGER.OH%d.DEBUG_LAST_CLUSTER_%d",(int)oh,(int)cs));
+  return readReg("GEM_AMC."+toolbox::toString("TRIGGER.OH%d.DEBUG_LAST_CLUSTER_%d",(int)oh,(int)cs));
 }
 
-uint32_t gem::hw::HwGenericAMC::getOptoHybridTriggerLinkCount(uint8_t const& oh, uint8_t const& link, AMCOHLinkCountT const& count)
+uint32_t gem::hw::amc::HwGenericAMC::getOptoHybridTriggerLinkCount(uint8_t const& oh, uint8_t const& link, AMCOHLinkCountT const& count)
```
jsturdy, line 698: ~~likewise here~~ i guess not, as i see this one was *not* ported over...

gemhardware/managers/Makefile
```diff
@@ -47,6 +51,7 @@ IncludeDirs+=$(BUILD_HOME)/gembase/include
 IncludeDirs+=$(BUILD_HOME)/gemreadout/include
 IncludeDirs+=$(uHALROOT)/include
```
jsturdy, line 52: probably don't need

gemhardware/managers/src/common/optohybrid/OptoHybridManager.cc
```diff
@@ -360,81 +191,32 @@ void gem::hw::optohybrid::OptoHybridManager::configureAction()
...
-
     for (size_t link = 0; link < MAX_OPTOHYBRIDS_PER_AMC; ++link) {
       size_t index = (slot*MAX_OPTOHYBRIDS_PER_AMC)+link;
       CMSGEMOS_DEBUG("OptoHybridManager::index = " << index);
       OptoHybridInfo& info = m_optohybridInfo[index].bag;
-
       if (!info.present)
         continue;
-
+
```
jsturdy: clean me!

jsturdy (dismissed):

mexanick:
**Test update**
With a cool enough front end all the FSM transitions are executed as before

gemhardware/managers/src/common/amc/AMCManager.cc
mexanick: I'd say there's more to the calibration suite tests... The system also can't be easily recovered from cold boot state. But if you insist, I can retrofit this back. However I can't test the real data taking until we fix the cooling

gemhardware/devices/Makefile
mexanick: uhal is still a dependency for the AMC13-related code

gemhardware/devices/src/common/amc/HwGenericAMC.cc
mexanick: basically it is a reference for future continued refactoring

gemhardware/devices/Makefile
jsturdy: indeed, but there should be *no* `amc13` code in `gemhardware/devices` (only in `gemhardware/managers`, no?)

gemhardware/devices/src/common/amc/HwGenericAMC.cc
mexanick: yes

gemhardware/devices/src/common/amc/HwGenericAMC.cc
mexanick: yes, but this is out of the scope of the current PR

gemhardware/managers/Makefile
mexanick: still need as AMC13Manager uses `uhal`

gemhardware/devices/include/gem/hw/devices/amc/HwGenericAMC.h
mexanick: done

gemhardware/devices/include/gem/hw/devices/amc/HwGenericAMC.h
mexanick: done

gemhardware/devices/Makefile
mexanick: indeed, you're right. Done.

lpetre-ulb (approved):
Approved as a technical PR. I prefer to see all new developments use the templated RPC and run into easily fixable issues rather than waiting months for the perfect PR; all issues from such a large change cannot be discovered by simple testing.

